PR #223, opened 2023-02-25
Translated parts_list.search.searching_for

sascha988:
Translated english text string parts_list.search.searching_for into german.


jbtronics:
Looks good. Thanks

PR #225, opened 2023-02-25
vulnerability XSS fix

sascha988:
The "trans with" command is not automatically escaping the string, so this is a XSS (Cross-Site Scripting) vulnerability. Tested string: https://URL-TO-PART-DB-SERVER/de/parts/search?keyword=%22'%3E%3Cqss%20a%3D X147208852Y1_1Z%3E

QUALYS Enterprise WAS Scan Report classifies this as level 5 security risk


jbtronics:
Thank you very much. Similar seems to be possible at other locations too. I will fix that there too.

PR #227, opened 2023-02-25
vulnerability Path-relative stylesheet import (PRSSI) fix

sascha988:
Threat
Relative URLs can be dangerous since browser may not determine the correct directory. If the HTML uses path-relative CSS links, it may be susceptible to pathrelative stylesheet import (PRSSI) vulnerabilities. This could allow an attacker to take advantage of CSS imports with relative URLs by overwriting their target file.

Impact
An attacker may trick browsers into importing JavaScript or HTML code as a stylesheet. This has been shown to enable a number of different attacks, including cross-site scripting (XSS) and exfiltration of CSRF tokens.

Solution
It is recommended to use absolute URLs for CSS imports. Alternately you can add the HTML "base" tag in the document which defines the base URL or target location for all the relative URLs.
The vulnerability can also be mitigated by using the following best practices to harden the web pages: • Set a DOCTYPE which does not allow Quirks mode as explained at https://hsivonen.fi/doctype/ • Set response header X-Frame-Options: deny
• Set response header X-Content-Type-Options: nosniff.

-----------
To me the easiest way to fix this, was adding the base URL. :)

### DIFF
--- a/templates/base.html.twig
+++ b/templates/base.html.twig
@@ -3,6 +3,8 @@
 <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+    
+    <base href="{{ app.request.getSchemeAndHttpHost() }}">
 
     <link rel="manifest" href="{{ asset('manifest.json') }}">
 


jbtronics:
Do you have an example where relative paths for stylesheets are used? Stylesheets are managed using webpack which should only generate absolute pathes I think.
The X-Fame-Options: deny header is already set.

Besides that I see no problem in adding the base tag. I will do some tests however, to ensure that locale prefixes and non path rewrite case are handled correctly.

sascha988:
Hello Jan,

You can use the `.setPublicPath('https://your-domain.as-absolute-path.to/build')` from Encore in webpack.config.js to force absolute URLs.
But if you're server has multiple names, then this is not a way of handling this...

For example in your [https://part-db.herokuapp.com/de/](url) Demo you can see in the source code the following line
`<link rel="stylesheet" href="/build/4977.b066e685.css" data-turbo-track="reload" integrity="sha384-OF2c859t9umk6RNjAIhrfgco1sqVMyvYNudcpC7oIUr15SBRzaOCA1V/ZYgTDGNB">`

so there is a relative URL to `/build/4977.b066e685.css`

jbtronics:
As far as I understand, this is an absolute path with the leading path (they will always have point to an absolute path on the current server). In general, I think the PRSSI attack should be nearly impossible in Part-DB as the only non-assets files which can be called from the outside is the symfony frontend controller, which errors when encountering unknown pathes. So using this for an PRSSI attack should be extremely difficult.

The problem with the base tag is that it also applies to anchor links, which should normally just jump to the element on the current page. With the base tag set, these would all point to the homepage, which breaks some things. 

This is maybe fixable with some bigger changes, but for the moment I will not merge this, as this would break stuff.

PR #294, opened 2023-05-27
docker: add missing PassEnv directives

japm48:
There were some missing variables in the Apache docker configuration.

In particular, it took me *ages* of debugging to understand that `TRUSTED_PROXIES` was not being passed to php-fpm, which rendered my HTTPS reverse proxy setup useless.


jbtronics:
looks good. Thank you

Issue #360, opened 2023-09-03
Allow configuring default columns and their order using environment variable.


misaz:
By default there are some columns preselected. I would like to change my server-wide defaults. I would like to have option for configuring

1) enabled columns
2) their order

for example:

```
PART_TABLE_DEFAULT_COLUMNS=ipn,edit,mpn,description
```

As far as I understood sources, it will require extending [PartsDataTable.php](https://github.com/Part-DB/Part-DB-server/blob/master/src/DataTables/PartsDataTable.php). Maybe some further refactoring will be needed. Currently it has hardwired defaults and order.

This can apply to other tables also, but I personaly "need" it only for parts table.

![Default columns](https://github.com/Part-DB/Part-DB-server/assets/2937706/b5a94285-1f86-4e71-96ce-20b690828f55)

jbtronics:
This is fixed with PR #390

PR #390, opened 2023-10-04
Support for configuring default columns in part table

misaz:
I implemented myself the feature which I mentioned in issue https://github.com/Part-DB/Part-DB-server/issues/360 and I would like to contribute it to the mainline (master). Current implementation is quite a simple, but whole method `configure` seems too long to me (It was alredy too long and combined too much logic at the begining and now it is even slightly worse). In my opinion it would be better to do refactoring and add additional layer to abstraction of columns by separating every column to dedicated class and in PartsDataTable just instantiate them instead of manually configuring each column. Also outsorcing sorting and visibility enabling logic to separate class would be welcome, I think. I tried to refactor it, but it started to seems bit overcomplicated, especily when implementing DI instantiation. So I reverted to more simple solution for now. I considering contribute also possibilty to configure (per user/category) parameters which will show in the table. In this case I would do the refactoring.

Could you please check that my contribution is
a) correct?
b) satisfy Part DB codebase requirements and all other requirements?

and possibly marge it to master?

Thanks.


jbtronics:
Hey, thank you for your contribution.

I have moved the ordering logic to a separate service, to improve the DX and making it more flexible and easier to use on other tables.
I changed the env option name and some of the column names slightly. The feature is merged.

To configure that on a per user base, this should be utilizing the datatables frontend mechanism. Currently you can change the visibility and order of columns in the frontend, but this is just saved in the browser. Basically this just have to be saved in the user profile in Part-DBs database, and the user can change that easily using datatables features in the browser.

PR #421, opened 2023-10-25
Respect the price related quantity if reported by the InfoProvider

d-buchmann:
It seems like all InfoProviders so far reported their parts prices in multiples of 1.
This adds the price related quantity as an option to the PriceDTO.
(I needed this for my EttingerInfoProvider I want to add as a PR, but their API is yet in beta stage)


jbtronics:
Makes sense. Thank you for the PR.

PR #433, opened 2023-11-16
Associate project BOM entries with correct project when cloning

### LINKED ISSUE #432: Cloning a project puts the BOM clone into the wrong place
**Describe the bug**
Cloning a project duplicates the BOM in the original project instead of putting it into the newly created one.

**To Reproduce**
1. Create a project with a BOM.
2. Click on 'duplicate element'
3. Review the BOM, save the project under a different name
4. Review both projects: The original has twice the BOM entries, the new one has none.

**Expected behavior**
The new BOM entries are created, but associated to the wrong (=original) project.
Instead, they should be associated to the dupicated project.

**Server Side**
- Part-DB Version: 1.8.1
- PHP Version: 8.2.11
- Database Server: 10.3.38-MariaDB-0ubuntu0.20.04.1

**Additional context** 
May or may not be related to #302.

d-buchmann:
This fixes #432.

### DIFF
--- a/src/Entity/ProjectSystem/Project.php
+++ b/src/Entity/ProjectSystem/Project.php
@@ -186,7 +186,7 @@ class Project extends AbstractStructuralDBElement
             //Set master attachment is needed
             foreach ($bom_entries as $bom_entry) {
                 $clone = clone $bom_entry;
-                $this->bom_entries->add($clone);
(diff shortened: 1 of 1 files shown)


jbtronics:
Thank you.

Issue #90, opened 2021-01-06
Parameter Text field should handle text same as Symbol field


fk0815:
Currently the text is not rendered as for the symbol field. With this feature added test condiitons like "Ic = 500mA, Vce = 1V" could be written more beautiful.

PR #463, opened 2023-12-22
Update helper.twig

Xernito:
Better display of units so that superscript and subscript values can be used.

See Issues #90


Xernito:
See Issues #90

jbtronics:
I don't think this implementation is ideal, as it requires to manually wrap the unit with `$` to render it as latex equation. This should be done automatically (this should be possible by creating a new functions on the parameters, which returns a formatted version of the value with the latex wrapped unit.

Besides that units normally are type set using normal (non italic) fonts, while katex use an italic font in expressions by default.
So the unit should also get wrapped in a `\mathrm` statement to get the right font.

Xernito:
Okay, i will look if I can do a better implementation.

PR #526, opened 2024-02-22
Add LCSC Info Provider using their public API

meganukebmp:
Recently at the lab I work at, we needed an inventory management system. PartDB fit that role very nicely, great software btw. However a lot of our components for the various projects we work on came from the LCSC supplier. As such we had to find them on the other info suppliers manually and it was taking a lot of time. Our current workflow involves scanning the product tag with a barcode scanner and obtaining the vendor-supplied product id from there and plugging it into the info supplier search. LCSC has some components that we couldn't find in the other suppliers and also would have taken too long to manually import the thousands of parts we needed, as such we took the effort to implement LCSC's public API.

This API is used by the current LCSC frontend (javascript) to query their backend and then populate the UI. As it currently stands their official javascript frontend makes MORE API calls per search (it does a call for every character entered into the searchbox) than this PartDB implementation.

Currently all data fields are obtainable from LCSC except for part availability (if it's discontinued or whatnot) since LCSC does not provide this information on their storefront either. This includes pictures, datasheets, footprints, parameters and more.

Both plaintext and LCSC product numbers are supported.

Examples:
image
image
image
image

### DIFF
--- a/src/Services/InfoProviderSystem/Providers/LCSCProvider.php
+++ b/src/Services/InfoProviderSystem/Providers/LCSCProvider.php
@@ -1,0 +1,301 @@
+<?php
+/*
+ * This file is part of Part-DB (https://github.com/Part-DB/Part-DB-symfony).
+ *
+ *  Copyright (C) 2019 - 2023 Jan Böhmer (https://github.com/jbtronics)
+ *  Copyright (C) 2024 Nexrem (https://github.com/meganukebmp)
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU Affero General Public License as published
+ *  by the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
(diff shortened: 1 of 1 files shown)


jbtronics:
Thank you. I will try it out later. But seems good from a first glimpse.

However I will add some configuration option, to enable/disable this provider, so that administrators can choose which part providers they wanna use.

jbtronics:
I have added this configuration possibility and improved the parameter parsing, so that values like `4V~6V` get parsed as filterable  numeric range instead of a simple text value.

jbtronics:
I have also implemented a possibility to configure the currency to retrieve the price informations from LCSC in (possible via passing a cookie value).

Whats not fully clear is the discount value and whether it should be used for final price calculations. The online shop shows lower prices then the productPrice field from the API. The real value can be calculated using the given discount value.

Im not sure if these discounts are permenant and part of a marketing strategy or just temporary. In the first case it might be a good idea, to use the discounted value as price info.

PR #530, opened 2024-02-23
Fix RegEx to handle negative values and Ohms without prefix

frank-f:
This is a minor tweak for #526, specifically for https://github.com/Part-DB/Part-DB-server/issues/526#issuecomment-1960444586

I noticed, that (LCSC) attributes of ranges with negative values were still saved as text. That was due to the minus character, that was missing from the regular expression in splitIntoValueAndUnit().


frank-f:
I just noticed, that a parameter value of 120Ω was also not correctly parsed.

frank-f:
I'll add more tweaks while polishing my database, but won't explain each one. Just ask here, if you have questions.

jbtronics:
Looks great. Thank you.

PR #566, opened 2024-03-12
install sudo to be able to do db migrations from the container's console

au-ee:
We have Part-DB running containerized as a K8s deployment. 
After an update, we connect to the server pod, install sudo and do the db migration as `www-data`. 
We plan to do the migration with an init container in the future, but until we manage to do that, it would be great if you accepted to include `sudo` into the container image. 

Thanks!


jbtronics:
Hmm I would think that you should be able to execute a shell in the containers under another user from the host system. But I dont see anything speaking against including sudo, so thats fine I think.

The new docker image will probably contain automatic database migrations, but I need to figure out a good mechanism to perferm backups beforehands.

au-ee:
With Kuberentes/kubectl, there is no builtin way of attaching to a pod as a different user. You could use the container engine's tools but for that you would have to be logged in on the node currently running the workload.

Thanks for merging!

PR #582, opened 2024-03-30
LCSC: Follow first 'pdfUrl' link to get real datasheet URL

frank-f:
LCSC changed their website. The "pdfUrl" key does not contain the link to the actual PDF anymore. It now links to a website that shows the datasheet side by side with an "add to cart" form.
That does not work at all if it's downloaded by Part-DB so I patched the LCSC provider, to open the first link and extract the actual PDF URL from it. I was getting some strange errors while testing, so I included a Referrer and User-Agent header. It's been working 100% for me since then.

PS:
Unfortunately I could not find an API method to get the real datasheet URL, so I resorted to scraping the HTML website. If anyone can get their hands on an LCSC API description (official or not), I'll be more than happy to update the code.

### DIFF
--- a/src/Services/InfoProviderSystem/Providers/LCSCProvider.php
+++ b/src/Services/InfoProviderSystem/Providers/LCSCProvider.php
@@ -99,7 +99,6 @@ class LCSCProvider implements InfoProviderInterface
         if (!empty($url) && preg_match("/^https:\/\/(datasheet\.lcsc\.com|www\.lcsc\.com\/datasheet)\/.*(C\d+)\.pdf$/", $url, $matches) > 0) {
           $response = $this->lcscClient->request('GET', $url, [
               'headers' => [
-                  'User-agent' => 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.4.1 Safari/605.1.15',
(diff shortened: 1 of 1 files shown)


jbtronics:
See the review.
Besides it looks fine, I think. Thank you

jbtronics:
Its not about the Codecov reports.
Its about my comment about Line 107 and 108:

> This seems unnecessarily complicated. I think you could just do something like $url = trim($matches[2], '"'); to get rid of the quotes around the URL, instead of constructing a JSON object and decoding it immedeatly.

frank-f:
> Its not about the Codecov reports. Its about my comment about Line 107 and 108:
> 
> > This seems unnecessarily complicated. I think you could just do something like $url = trim($matches[2], '"'); to get rid of the quotes around the URL, instead of constructing a JSON object and decoding it immedeatly.

Oh, I don't see that comment anywhere.

I took this complicated looking approach, because the String is actually encoded and this seemed like the most fail-proof way to decode it.
It looks like this in the source:

`pdfUrl: "https:\u002F\u002Fwmsc.lcsc.com\u002Fwmsc\u002Fupload\u002Ffile\u002Fpdf\u002Fv2\u002Flcsc\u002F2304140030_Microchip-Tech-MCP2562T-E-MF_C191416.pdf",`

PR #583, opened 2024-03-31
Improve parameter parsing

frank-f:
This will improve the following cases (example values):

`-40°C ~ 125°C`

This worked for LCSC only. Other providers needed "..." as range indicator and values with units could not be processed at all.

`-40°C ~ 125°C @Tj`

Extra info at the end stopped the range detection altogether, because the two "units" would not match. Extra info will now be saved to the "Text" column, making the units match and thus allowing the actual values to be saved as range.

`22 µH`

"µ" was missing in the unit detection regex


frank-f:
Got all tests to pass, so from my point of view my patch ready to be merged.

Let me know, if you'd like anything changed.

jbtronics:
This seems like a good addition. I have added some tests for your code, so that changes dont break something unnoticed. Also fixed some edgecases.

Thank you

PR #596, opened 2024-04-16
Update Measurement Unit concept

mdiepart:
This small pull request adds a few explanations about the measurement unit parameters. Helps avoiding confusion about their presence and their role. It also makes it clear that those parameters are not inherited by the parts.


jbtronics:
Thank you.

PR #612, opened 2024-05-12
Update LCSC API URL

### LINKED ISSUE #611: LCSC Search URL outdated
**Describe the bug**
When using "Create parts from info provider" with LCSC, the search now fails with an error message.
I was able to successfully use the LCSC provider to search for parts a few days ago.

**To Reproduce**
1. Enable LCSC search using `PROVIDER_LCSC_ENABLED=1`
2. Go to 'Tools -> Create parts from info provider'
3. Input a search term and select 'LCSC' as the provider
4. Click search
5. See error

**Expected behavior**
Get search results from LCSC

**Screenshots**
image

**Server Side**
- Part-DB Version: 1.12.0 
- PHP Version: 8.1.28 (SAPI: fpm-fcgi)
- Database Server: SQLite 3.34.1

**Additional context**
By looking at the network trace and searching on LCSC.com, it looks like the global search base url changed from /**wmsc**/search/global to /**ftps/wm**/search/global.

SapuSeven:
This directly fixes #611 

It would still be nice to be able to customize the URL directly using an environment variable, something like `PROVIDER_LCSC_ENDPOINT_URL`.

### DIFF
--- a/src/Services/InfoProviderSystem/Providers/LCSCProvider.php
+++ b/src/Services/InfoProviderSystem/Providers/LCSCProvider.php
@@ -35,7 +35,7 @@ use Symfony\Contracts\HttpClient\HttpClientInterface;
 class LCSCProvider implements InfoProviderInterface
 {
 
-    private const ENDPOINT_URL = 'https://wmsc.lcsc.com/wmsc';
+    private const ENDPOINT_URL = 'https://wmsc.lcsc.com/ftps/wm';
 
     public const DISTRIBUTOR_NAME = 'LCSC';
 


SapuSeven:
Confirmed working with a local build:
image

jbtronics:
I'm not sure if an env would be a good idea long term. Because I would expect that the url is not the only thing that would change over time, and then you would need to change the code anyway.

jbtronics:
Thanks for your PR.

Split7fire:
Any plans to make dot-release?

Issue #621, opened 2024-05-26
add scanner and search buttons to the navbar in the mobile view


RaptorDE:
In the mobile view, the Scanner and Search buttons are moved to the hamburger menu. As these functions are frequently used on smartphones, I recommend always displaying them as buttons in the nav bar.

![2024-05-26 11_19_45-Part-DB – Mozilla Firefox](https://github.com/Part-DB/Part-DB-server/assets/37591931/0a785504-6688-45e9-a28d-e055bdabfac9)

jbtronics:
Fixed with the merge of #622

Issue #620, opened 2024-05-26
The “Show/Hide sidebar” button is invisible in dark mode


RaptorDE:
The “Show/Hide sidebar” button is invisible in dark mode because it has the same color as the background.

![Show-Hide sidebar](https://github.com/Part-DB/Part-DB-server/assets/37591931/e737e5d5-0a18-4aa7-8d55-ecbc163bdf92)

Instead of just changing the color, I suggest moving the button to the navbar, as in the mobile view. As the small button on the side can easily be overlooked and is very difficult to click in desktop mode on a smartphone.

Here is an example of how it could look:

![2024-05-26 11_02_04-Part-DB – Mozilla Firefox](https://github.com/Part-DB/Part-DB-server/assets/37591931/ed92e10d-cfe8-4810-b5db-a125b31abadb)

**Server Side:**
- https://demo.part-db.de/en/
- Part-DB Version: 1.12.0
- PHP Version: 8.2.18
- Database Server 10.4.32-MariaDB

**Desktop:**
 - OS: Windows 11
 - Browser Firefox
 - Version 126.0

**Smartphone:**
 - Device: Sasung S10
 - OS: Android 12
 - Browser Chrome
 - Version 125.0.6422.72

jbtronics:
fixed.

PR #622, opened 2024-05-26
added scan button and show/hide sidebar button to navbar

RaptorDE:
This is a suggestion regarding the issues #620 und #621 . This is my first attempt at Symfony and Bootstrap. So I am not sure if I have chosen the right approach.

The ‘Show/Hide sidebar’ button certainly has room for improvement. Currently both buttons are available but not synchronised. I haven't fixed the problem with the visibility in dark mode either. Maybe the original button can also be removed? I have also not yet touched the function that changes the icon of the button. That's why the arrow is still the icon.

I don't expect my pull request to be accepted without changes.

This is how it looks:
2024-05-26 15_40_59-Part-DB – Mozilla Firefox

sidebar-ezgif com-optimize


jbtronics:
I am not a fan of the sidebar collapse button in the navbar. I think it's too dominant there, and the function is more clear, when it is right next to the sidebar, like before.

I fixed the darkmode behavior of the sidebar collapse button, so that is not a problem anymore

RaptorDE:
Thank you for the bugfix, I have moved the scan button in the mobile view to a new pull request. #626

PR #623, opened 2024-05-26
the preview image size in tables has been adjusted for better visibility

RaptorDE:
before
Screenshot_small

after
Screenshot_big

### DIFF
--- a/assets/css/app/images.css
+++ b/assets/css/app/images.css
@@ -18,8 +18,8 @@
  */
 
 .hoverpic {
-    min-width: 10px;
-    max-width: 30px;
+    min-width: 30px;
+    max-width: 80px;
     display: block;
     margin-left: auto;
     margin-right: auto;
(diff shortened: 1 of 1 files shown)


jbtronics:
Honestly, I find the images too big that way and the tables become very long unnecessarily.

image

We can build in an option to have larger preview images, but that should be configurable. And for that it is probably the best to wait for the new settings system, I am working on.

RaptorDE:
I understand your point that the table becomes too long with too large a image. The idea of changing the image size via the settings is great.

But I think the min-width of 10px is too small, you can't see anything on the small preview images. I noticed this especially on my smartphone (Samsung S10 with Chrome).

Personally, I think a width of 30px is nicer.

This is what it looks like with 10px:
Screenshot_20240609_115912_Chrome

This is what it looks like with 30px:
Screenshot_20240609_115816_Chrome

jbtronics:
Now obsolete. See #984

PR #670, opened 2024-08-20
Update LCSCProvider field for real datasheet URL

frank-f:
I noticed, that LCSC changed the field containing the URL to the real datasheet.

I updated the LCSCProvider method accordingly.


jbtronics:
Thank you

PR #691, opened 2024-09-05
Implement order details

### LINKED ISSUE #217: Marking ordered components "waiting to arrive"
First of all, I want to thank you for the good work. I really like the direction this project is taking.

Right now I've created an inventory "Ordered", but this is not very convenient, because these parts are added to the total number of parts available. It would also be useful to be able to show the entire list of components that are already being delivered. 

In addition, I would like to see more information about the parts amounts on the project's build page under the "Build" tab. To more clearly understand how much I have in stock and how much I need to order.

If you paid attention to these aspects, it would be just wonderful!

fabi321:
This adds the possibility to add information about current orders to items.

I'm not sure, how to properly handle translations, and sqlite backwards migrations.

Fixes #217


jbtronics:
I do not thing it is really useful to track orders with a simple number at part level, the old Part-DB versions did something similar (with the still existing `order_quantity` field), but it was quite limited and not so useful for many applications.

I think a order-centered system would be more flexibile and useful. Users can create Orders (as a new entity), assign parts and amounts to it and track its status via it (maybe you could even allow to order directly from Part-DB via the distributors APIs). That way you can have multiple orders of the same part in parallel (without loosing detailed track), can store metadata (like who ordered, and when) and when the order arrives you can easily book in all parts of this order, without needing to do that for each part at once.

This also allows to look for historic orders and other things, which should be useful, and which would not be possible by simply storing the currently in order state amount of parts.

barisgit:
Would this still be a useful feature and is someone working on it already? Also one thing that goes well with it here is this: Export BOM from KiCad/Altium and then import it in here and select how many times of components we need and then only export BOM for supplier (LCSC, Mouser, Digikey...) to buy only what you need.

PR #755, opened 2024-11-01
Responsive tweaks

dotsam:
Some tweaks to better support mobile widths. The diff here doesn't look great owing to some larger markup/nesting changes, but here's what's happening:

- Tweak the data tables layout, as it appears it was using some Bootstrap 4 markup that's no longer valid
- Tweak the single part info, mainly making the image full-width and smaller screen sizes so that the image slider is usable and therefore the other two columns have more width to display in
- Added some more checks around part properties so no empty markup is output
- Converted the displayed part properties into a `<dl>` Description List and removed the use of `<h6>`, as this is a better semantic fit. Also provided more accessible markup while retaining the tooltips

If this final change to `<dl>`'s looks okay to you, I'm also working on doing the same thing to the single part "sidebar" and can submit a PR for that too

### DIFF
--- a/config/packages/datatables.yaml
+++ b/config/packages/datatables.yaml
@@ -10,13 +10,12 @@ datatables:
     options:
         lengthMenu : [[10, 25, 50, 100, -1], [10, 25, 50, 100, "All"]]
         pageLength: '%partdb.table.default_page_size%'   # Set to -1 to disable pagination (i.e. show all rows) by default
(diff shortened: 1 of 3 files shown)


jbtronics:
looks good, thank you.
I will just adjust the font-weight of the <dl> elements to match the <h6> style, as they should appear slightly highlighted over normal text.

PR #794, opened 2024-12-07
A new handheld scanner function

barclaac:
This adds a prototype handheld scanner support to support EIGP114 barcodes (like Mouser and Digikey use.) Enables a very simple workflow of:
1) Scan container
2) Scan package (with potential autocommit)
3) Adjust quantity if needed
4) Submit
5) Goto #2

This gives a very efficient workflow for checking in an order when it arrives.

This is still very rough with zero translations done and certainly not ready for prime time. Feedback welcome, I'm not a php or js developer.


jbtronics:
That is implemented in recent Part-DB releases.

PR #810, opened 2024-12-21
Show when parts from info provider already exist

Treeed:
When adding the new part via the info provider system, the system would make a new part, regardless of the identical old one. I changed the info provider service to look for each part that is listed in the info provider search results and add a button to view the existing part if one exists. 

**Translations:**
No idea how to add those, but needs to be done

**Style:** 
Currently the layout is functional, but I would like some clearer notice that the buttons are there because the part was found locally. Because the row is already quite full and there's not much space to add info I would ideally have liked it to look something like this:
image
(Screenshot of a random webpage for illustration purposes) 
I'm not great at html/css though and adding a dummy row doesn't play well with the striped layout of the table.

### DIFF
--- a/src/Controller/InfoProviderController.php
+++ b/src/Controller/InfoProviderController.php
@@ -75,31 +75,31 @@ class InfoProviderController extends  AbstractController
     private function matchResultsToKnownParts(array $partsList): array
     {
         //we need a manufacturer object to look for a manufacturer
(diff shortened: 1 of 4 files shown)


dm5tt:
Upvoting this. Added and ordered (lol) the same components over and over again. This would be a great addition.

jbtronics:
Thank you for your PR again.

I have improved the visual style and added translations. Now it looks like this:

image

I have optimized the query logic to be faster and find the results more precisely.

PR #811, opened 2024-12-21
Added capability to scan Digikey barcodes and open the local part part page based on the result

Treeed:
All of the digikey bags already have barcodes with all the information needed to find a part, so registering the barcodes individually doesn't seem so useful. I added a decoder for the digikey barcodes which allows directly jumping to the part without registering the barcode. It should be noted, that the current implementation of the barcode scanner is hardly able to scan Digikey's Datamatrix codes, so more work needs to be done to make this pull request useful.
See this discussion for more info: https://github.com/Part-DB/Part-DB-server/discussions/808

This feature has a bit of an issue naming wise. The existing implementation already defines vendor barcodes as user defined ids that can be added as a way to identify a lot (and find that lot by scanning a barcode with that id). I find that naming very confusing, because the way the feature works doesn't care about whether this is a vendor barcode or anything else that you're adding; it's just any user defined scanned string that you want to use. When I first found that feature it took me a good while to understand what it actually does, especially since I could find any documentation on it and I didn't manage to connect the name "part lot" to "part stock". 
Since this pull request adds support for actually reading and extracting the information from, specifically, vendor barcodes I felt it would be best to rename the old feature to "user defined barcodes". I found however, that the naming spans across so many files, that the merit is questionable. In the files I committed I started this renaming process but didn't complete it across all usages of the term "vendor" in properties and variables. (variable names for the same variable are consistent, i.e. no functionality is broken, only the naming among different variables or properties is not consistent)
Generally, introducing barcodes that actually contain information is rather different from the old concept where a barcode was just an identifier. I have created a new data class VendorBarcodeScanResult, but that means some of the functions now have to handle two different kinds of objects. Maybe it makes sense to separate concerns further and break the interpretation of id barcodes and information barcodes up into two classes. Also, in contrast to id barcodes, vendor barcode formats should be easy to add. It might make sense to create a list of known formats and corresponding parser functions somewhere and use that to generate the options in the ui and try out the different parsers

### DIFF
--- a/src/Controller/ScanController.php
+++ b/src/Controller/ScanController.php
@@ -42,10 +42,10 @@ declare(strict_types=1);
 namespace App\Controller;
 
 use App\Form\LabelSystem\ScanDialogType;
-use App\Services\LabelSystem\Barcodes\BarcodeScanHelper;
-use App\Services\LabelSystem\Barcodes\BarcodeRedirector;
-use App\Services\LabelSystem\Barcodes\LocalBarcodeScanResult;
-use App\Services\LabelSystem\Barcodes\BarcodeSourceType;
+use App\Services\LabelSystem\BarcodeScanner\BarcodeRedirector;
+use App\Services\LabelSystem\BarcodeScanner\BarcodeScanHelper;
+use App\Services\LabelSystem\BarcodeScanner\BarcodeSourceType;
+use App\Services\LabelSystem\BarcodeScanner\LocalBarcodeScanResult;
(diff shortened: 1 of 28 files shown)


jbtronics:
Thank you.
I have extracted the EIGP114 decoding into its own data class, so that you can access the label info in a typesafe way, and hopefully be able to reuse that info in other places.

Also I added a little option to just list all the decoded content of a barcode in the scanner tool: 

image

PR #847, opened 2025-02-02
Fixed attachment statistics for sqlite

Treeed:
SQLite doesn't support the backslash as an escape character by default. Any escape character must be specified by the ESCAPE function. To avoid confusion when using other DBs which do recognize the backslash, but also the ESCAPE function I used # here.


jbtronics:
Looks good. Thank you.

PR #848, opened 2025-02-02
Split attachment paths

Treeed:
Since there is a tendency for online data to be lost it usually makes sense to generally download any external attachments. However, with the current configuration, the original link is lost, so there is no easy way to check e.g. if there was a change to the original file. Since I don't like throwing out data in bulk I needed a mechanism to retain both copies.

This PR splits the single "path" property into "internal_path" and "external_path" to achieve that. To keep the naming and handling of internal and external files consistent that spreads into quite a few files. 

Sorry for the single huge commit, but there's really no in-between state where everything still works.

Some translations are required again. Since I still don't know how to do these I just added TODO tags. There's also some changes to the English translation file which would need to be propagated to the other languages.

Depends on PR https://github.com/Part-DB/Part-DB-server/pull/847


jbtronics:
This is a useful and reasonable change, I think. Thank you for the PR.

I am currently going through it, add the missing translations and change some smaller things.
I made the external_path and internal_path columns nullable in the DB and use null as empty value. In my opinion, this is more logical and consistent with other empty value definitions.

Also, I think we should keep the media_url property in the serialized API output for backward compatibility (while removing its documentation, so that it is not used for newer software). That should be an easy addition.

PR #849, opened 2025-02-02
Highlightable attachment paths

Treeed:
When editing a part there is no way to highlight an attachment path in order to copy it. There is also no way to edit it without completely removing it, even if it has just been entered.
Another nuisance is that when entering a path that has been in that field before, that text is lost when clicking outside the field. 

I wrote two small Tom Select plugins to fix that. 

click_to_edit would probably also be nice to have in any other Tom Select fields that allow creating options and autoselect_typed would probably be nice to have in any other Tom Select fields in general.


jbtronics:
Looks good. Thanks

PR #873, opened 2025-02-25
Update LCSCProvider.php: fix error in query string

mkne:
Fix typo in query string 'prodctCode' -> 'productCode' introduced in Commit 80527e3


jbtronics:
Thanks you

PR #875, opened 2025-02-27
Modified the DigiKey Provider to works with the V4 API

Danixu:
- Fixed a problem of unauthorized request using the V3 API by migrating to the V4.
- The new API returns the product variations in the same request which will be threated as different products.
- Fixed the Digikey provider functions to match the V4 format

### DIFF
--- a/src/Services/InfoProviderSystem/Providers/DigikeyProvider.php
+++ b/src/Services/InfoProviderSystem/Providers/DigikeyProvider.php
@@ -108,12 +108,13 @@ class DigikeyProvider implements InfoProviderInterface
     {
         $request = [
             'Keywords' => $keyword,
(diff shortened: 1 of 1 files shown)


Danixu:
My new developer account was not able to search on the v3 API, receiving all the time 401 unauthorized errors. I don't know if it's related with my account (it is a new account), or is just that they have finally removed the v3 API support. I have just modified the functions to match the v4 API returns.

rickyelopez:
> My new developer account was not able to search on the v3 API

I'm having the same problem. I only have the option to enable the v4 API in my production app settings page. I would love to be able to use this feature!

jbtronics:
Thank you, looks good.
I changed and improved two slight things.

PR #939, opened 2025-05-22
Update markdown_controller.js

d-buchmann:
Don't handle links as external by default. Instead distiguish internal (relative) and external (absolute) links.


jbtronics:
An internal path can be absolute too, but I see no disadvantage in having this change.

Issue #178, opened 2022-09-15
Edit or replace tags en masse.


virtadpt:
**Is your feature request related to a problem? Please describe.**
When I realize that I've misfiled or misspelled a tag name across my inventory (which happens when you clone entries to speed data entry) it takes a very long time to edit each and every item and correct the tag.

**Describe the solution you'd like**
It would be very helpful to have a function in Tools -> Tools where I could do a global search and replace on a tag (e.g., micocntrollers -> microcontrollers)

**Describe alternatives you've considered**
Search on the tag.  Open each hit in a separate tab.  Edit each hit in the separate tab to expose the tags.  Correct the tag.

**Additional context**
It would also be very helpful if the default component view could include the tags on a persistent basis.

jagauthier:
Can you drop into the database and just change it there?

virtadpt:
I could.  But isn't that killing a fly with a bazooka?

jagauthier:
I would consider bandaging a self-inflicted wound.

zanna-37:
I think that tags should be treated the same as categories, storage locations, etc.
It should be possible to edit, view and delete them in bulk.
One thing that I find frustrating is that when choosing the tags for a new part, I need to remember the tag name exactly, as the dropdown doesn't show them all.

PR #940, opened 2025-05-22
Allow bulk editing of tags

d-buchmann:
- Changed "target" parameter type of handleAction in parts table controller to string
- Sanitize input for existing actions which expect an int
- Add tag (dummy) to SelectAPI
- Add tag option to twig template

This is purely theory, nothing of this is tested yet!
When implemetation is done, this should solve #178.


mkne:
Hi @d-buchmann
any progress on this? I introduced part-db to a co-worker and we consider to use tags to label parts with RoHS and REACH compliance. So bulk editing of tags would come in handy for our existing 2000+ parts. Can I help with testing or something else?

d-buchmann:
Indeed I got a little further, but my lack of js skills is holding me back.
It feels like the solution is quite hacky all over the place =)
The worst thing is that all actions work with an "ID" of the target to be modified, but tags don't have one, so I abuse the ID parameter by putting in the tag verbatim.

I can add and remove tags by entering them in the selector field by hand. The search doesn't work yet.
If you try to remove a tag that doesn't exist, or click add when no tag is in the selector, nothing happens.
However this is only from my experience, not thoroughly tested. If you intend to test it, please do a full backup first :D
Note, this branch is still based on 1.x.

mkne:
> Indeed I got a little further, but my lack of js skills is holding me back. It feels like the solution is quite hacky all over the place =) The worst thing is that all actions work with an "ID" of the target to be modified, but tags don't have one, so I abuse the ID parameter by putting in the tag verbatim.
> 
> I can add and remove tags by entering them in the selector field by hand. The search doesn't work yet. If you try to remove a tag that doesn't exist, or click add when no tag is in the selector, nothing happens. However this is only from my experience, not thoroughly tested. If you intend to test it, please do a full backup first :D Note, this branch is still based on 1.x.

Sorry, my skills are even more limited here.  I only have a productive instance, so I can't easily test it. But the thing with the "ID" is interesting and fueled my curiosity. While looking into the database to understand your point I found out that it's currently easier for me to do it directly in database. So I used a SQL statement to add the tags for the parts that need it. But I still think your pull-request is useful. Thanks for your inspiration.

PR #944, opened 2025-05-28
More generous search

### LINKED ISSUE #892: Better search
When searching for a part named eg. "Display" > "Display" shows all results with display, but "Display " (with space) shows only results with the space. Also when searching for eg. display for K30 ("Display" in the name and "K30" in the description) the search won't look into description for the search words, only name OR description.

 When searching for keyworks combine results from part names and description

None

**Additional context**
None

d-buchmann:
Split the keyword into tokens (at spaces).
Search for every token individually, then combine results with AND.
Caution: This has no effect on the typeahead search, I will have a look at this soon.
Should fix #892.


jbtronics:
I wonder if it might be better to use the fulltext seach capabilities of the database then trying to implement more complex searches ourselves.

virtadpt:
@jbtronics Generally speaking, that's always a good idea.  Let the database server do the heavy lifting.

d-buchmann:
I've read another article about FTS (in MySQL specifically).
It stated "you cannot search across different tables by one index since you cannot add an index across different tables and a view cannot have an index".
Since we're potentially using many different tables to search in, I don't see an advantage of using FTS in this use case.
I'm not saying my made-up method is better, but it's db-agnostic and readily-available.
Anyway, I'm closing this PR in favor of a new one I have prepared, making the "advanced search" an optional setting.

I'd really appreciate if someone finds the time to review and test this new approach =)

PR #955, opened 2025-06-20
Add supplier information to KiCad part exports

barisgit:
- Include supplier name and part numbers from order details in KiCad exports
- Handle multiple suppliers with sequential numbering (Supplier 2, Supplier 3, etc.)
- Include both active and obsolete order details for comprehensive supplier info
- Add null checks to prevent errors when supplier or part number is missing


jbtronics:
Thank you.

PR #959, opened 2025-07-01
Update AbstractParameter.php

### LINKED ISSUE #958: Parameter value '0' not shown in parameter tab
**Describe the bug**
Parameter value "0" is not shown in the parameter view of a part.
Also the percent sign % for the unit of the Tolerance parameter is not correctly displayed. (It shows $\mathrm{%} instead)

**To Reproduce**
1. Add a 0R resistor with a parameter Resistance = 0 (or 0E0 or 0.0) 
(for example 0603WAF0000T5E from Info service provider LCSC)
2. Open the tab Parameter in the part details

**Expected behavior**
A parameter value of 0 Ohms is displayed in the Value column of the parameter table.

**Screenshots**

Image

**Server Side**
- Part-DB Version: 1.17.1
- PHP Version: 8.3.21
- Database Server: sqlite 3.40.1

**Desktop (please complete the following information):**
 - OS: Windows 11 
 - Browser: Chrome 
 - Version: 137.0.7151.122

**Additional context**
I have tried to manually change the value to 0 or 0.0 but it will always be converted to 0E0 which is of course also correct.

d-buchmann:
Make lazy null conditionals explicit, thus enabling output of `0` as parameter value.
Try to handle LaTeX special chars gracefully (especially %, as seen in the referenced issue).
Fixes #958

(Javascript code will need some work too: the LaTeX controllers also don't handle special chars.
The parameter selection drop-down show some params incorrectly, e.g.  `Tolerance [%]` shows as `Tolerance[`.
The parameter preview shows an error e.g. when entering `%`; With `\%`, it works.)

### DIFF
--- a/assets/controllers/pages/latex_preview_controller.js
+++ b/assets/controllers/pages/latex_preview_controller.js
@@ -33,7 +33,10 @@ export default class extends Controller {
     {
         let value = "";
         if (this.unitValue) {
-            value = "\\mathrm{" + this.inputTarget.value + "}";
+            //Escape percentage signs
+            value = this.inputTarget.value.replace(/%/g, '\\%');
+
+            value = "\\mathrm{" + value + "}";
         } else {
             value = this.inputTarget.value;
         }
--- a/assets/controllers/pages/parameters_autocomplete_controller.js
+++ b/assets/controllers/pages/parameters_autocomplete_controller.js
@@ -85,7 +85,9 @@ export default class extends Controller
                         tmp += '<span>' + katex.renderToString(data.symbol) + '</span>'
                     }
                     if (data.unit) {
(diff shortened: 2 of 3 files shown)


jbtronics:
I think it is not useful to escape all latex characters, as you then cannot use latex in units anymore at all (which was the whole point for using latex for formatting the units). For simple units you don't need it, but for something like $nV/\sqrt{Hz}$, latex is very useful.

I changed it in a way, that only the percentage sign gets escaped, as it is very unlikely that somebody wants to put a comment into a unit, and it will be more likely used as a percentage sign.

d-buchmann:
Make perfect sense.
I still have one small edit though: % inside the regex needn't be escaped, but if the user already escaped the percent sign, it would result in `\\%`, which we don't want.
Therefore capture and replace an optional backslash before the percentage sign as well.

PR #960, opened 2025-07-02
Add env option to disable part image overlay

### LINKED ISSUE #369: Can we please get rid of / disable the filename overlay?
Can the filename overlay on images please be removed / disabled?

image

When browsing the images the overlay makes me have to move my mouse out of the image (to get rid of the overlay), go back in  back on the arrow left/right, click, back out of the image (to get rid of overlay), repeat...

The overlay makes it near impossible to see the image.

d-buchmann:
Fixes #369 while preserving the state as-is


jbtronics:
Thank you.

PR #970, opened 2025-07-21
Adapted PartNormalizer to PHP 8.4.

monoxide13:
Was missing ? to allow null value.
PHP 8.4: Implicitly nullable parameter declarations deprecated


jbtronics:
This PR became obsolete with the upgrades to v2

PR #981, opened 2025-08-01
Feature/add xlsx support

barisgit:
- Allow using xlsx and xls to import parts
- Update example csv to better show what is possible
- Add required tests (PASS)
- Fix parts merger when using update from info supplier and consider parts that already have a spn populated


barisgit:
@jbtronics This PR can be closed right?

jbtronics:
If everything is contained in the other pr, then yes

barisgit:
I accidentally continued to work right on top of this branch, so unless you pushed something it should be good.

jbtronics:
Became obsolete by merge of #982

PR #982, opened 2025-08-02
Feature/batch info provider import

barisgit:
- Add ability to batch update parts from info providers

Works in the following way:
1. User selects parts from any table and chooses action 'Bulk import from info provider'
image

2. Gets redirected to step 1, where the user needs to choose how to search for parts (eg. user mpn to search on LCSC and MOUSER and at the same time use lcsc spn to search on LCSC)
image
image

4. Upon clicking "Search all parts", the server searches for all parts at once, presenting user with similar interface to single part import but for multiple parts, while at the same time creating a row in a new table 'bulk_info_provider_import_jobs' so that the progress is not lost upon refreshing. Optionally user can also check to prefetch details for every part in this stage (if the search returns too many parts, for example if we were to search by a vague keyword, this could mean it takes a long time, but can be prevented by choosing supplier part numbers instead). Prefetched data gets automatically stored in cache
image

6. The job is then available to be opened from tools pane as well. On the job pane the user is presented with a list of all parts, where each part has search results. When user clicks update part they get redirected to info provider edit, with additional query parameter that tells page it is associated with a job. This way once saving part there is a button to go back, but at the same time mark part as completed in a job.
image
image
image

---
Note: I have by a mistake continued to work right on top of my other addition (excel support), so it would be best to first finish that or maybe even discard it and do everything here.

### DIFF
--- a/src/Controller/BulkInfoProviderImportController.php
+++ b/src/Controller/BulkInfoProviderImportController.php
@@ -101,7 +101,7 @@ class BulkInfoProviderImportController extends AbstractController
         return $job;
     }
 
-    private function updatePartSearchResults(BulkInfoProviderImportJob $job, int $partId, ?BulkSearchPartResultsDTO $newResults): void
+    private function updatePartSearchResults(BulkInfoProviderImportJob $job, ?BulkSearchPartResultsDTO $newResults): void
     {
         if ($newResults === null) {
             return;
@@ -111,7 +111,7 @@ class BulkInfoProviderImportController extends AbstractController
         $allResults = $job->getSearchResults($this->entityManager);
 
         // Find and update the results for this specific part
-        $allResults = $allResults->replaceResultsForPart($partId, $newResults);
+        $allResults = $allResults->replaceResultsForPart($newResults);
 
         // Save updated results back to job
         $job->setSearchResults($allResults);
@@ -482,7 +482,7 @@ class BulkInfoProviderImportController extends AbstractController
             }
 
             // Update the job's search results for this specific part efficiently
-            $this->updatePartSearchResults($job, $partId, $searchResultsDto[0] ?? null);
+            $this->updatePartSearchResults($job, $searchResultsDto[0] ?? null);
 
             // Prefetch details if requested
             if ($prefetchDetails && $searchResultsDto !== null) {
--- a/src/Services/InfoProviderSystem/DTOs/BulkSearchResponseDTO.php
+++ b/src/Services/InfoProviderSystem/DTOs/BulkSearchResponseDTO.php
@@ -41,19 +41,25 @@ readonly class BulkSearchResponseDTO implements \ArrayAccess, \IteratorAggregate
 
     /**
      * Replaces the search results for a specific part, and returns a new instance.
-     * @param  Part|int  $part
+     * The part to replaced, is identified by the part property of the new_results parameter.
+     * The original instance remains unchanged.
      * @param  BulkSearchPartResultsDTO  $new_results
      * @return BulkSearchResponseDTO
      */
-    public function replaceResultsForPart(Part|int $part, BulkSearchPartResultsDTO $new_results): self
+    public function replaceResultsForPart(BulkSearchPartResultsDTO $new_results): self
     {
         $array = $this->partResults;
+        $replaced = false;
         foreach ($array as $index => $partResult) {
-            if (($part instanceof Part && $partResult->part->getId() === $part->getId()) ||
-                ($partResult->part->getId() === $part)) {
+            if ($partResult->part === $new_results->part) {
                 $array[$index] = $new_results;
+                $replaced = true;
                 break;
             }
+        }
+
+        if (!$replaced) {
+            throw new \InvalidArgumentException("Part not found in existing results.");
         }
 
         return new self($array);
--- a/tests/Services/InfoProviderSystem/DTOs/BulkSearchFieldMappingDTOTest.php
+++ b/tests/Services/InfoProviderSystem/DTOs/BulkSearchFieldMappingDTOTest.php
@@ -1,0 +1,63 @@
+<?php
+/*
+ * This file is part of Part-DB (https://github.com/Part-DB/Part-DB-symfony).
+ *
+ *  Copyright (C) 2019 - 2025 Jan Böhmer (https://github.com/jbtronics)
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU Affero General Public License as published
+ *  by the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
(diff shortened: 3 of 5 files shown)


barisgit:
Here is description of excel functionality pull report

- Allow using xlsx and xls to import parts
- Update example csv to better show what is possible
- Add required tests (PASS)
- Fix parts merger when using update from info supplier to consider parts that already have a spn populated a duplicate

barisgit:
Added a tab for filtering by job statuses and such
image

barisgit:
Do not merge yet, there is a bug when user doesn't check prefetch and it still fetches info, causing timeout on larger sets.

barisgit:
Add priority support in mapping table
image
image

barisgit:
Hey @jbtronics any clue when this could be merged/rejected? Otherwise we will need to continue on our own fork, not providing any contributions here anymore.

jbtronics:
I will try to release the V2 version in the next few days, it's basically only missing some documentation. Then I can review and merge the PRs.

As some base stuff changed, it will probably be easier to merge the PRs into the V2 changes than the other way around....

jbtronics:
I have started reviewing the PR. I have rebased it on the master branch with v2. As i wasn't sure if I am allowed to force push and overwrite your branch (or if you are using it for something else), there is now a branch with the same name in the main repo. If you have no problem with the force push, just say it, and i will do it (makes some PR features a bit easier i think).

I really like the feature, its nicely done in the WebUI.
I have already changed some few things related to the v2 changes so far and some smaller things related to the excel exports.

I am not finished with the review, but some remarks so far:
* The `BulkInfoProviderImportController` is really long and complex. I think functionality like the serialization/deserialization can be moved to the entities. Maybe introducing some DTOs instead of the complex arrays and move some of the logic to a service which is independent of the form handling would help too.
* I don't really like the special handling for the batch handling of LCSC. It is not really intrinsically unique to LCSC, and should be more universally available. I already created a new BatchInfoProvider interface, that if implemented make the searchKeywordsBatch method implementable by all info providers. The decision whether to use the batch or individual import should be based on this interface. Not hardcoded for LCSC
* The inline scripts in the templates will most likely not work in productive use, as then the Content-security-policy is active that will block this. Either you need to put it into a js file that will get included, a stimulus controller, or you need to add a nonce to the inline script tag. But i haven't tried out this so far and i don't know how good this work with turbo.

barisgit:
You may force push, I have my fork just to contribute, I have no plans to have own features on it.

- Do you want me to refactor BulkInfoProvider or will you do it in V2?
- Regarding hardcoded LCSC supplier number import I agree, but will BatchInfoProvider interface allow to search by supplier numbers? I haven't really tested with other than LCSC and Mouser as I don't regularly use others, but on them at least searching by MPN on batch input returns a very large number of results, which overloads API or in case of LCSC temporarily bans your ip, because it tries to then fetch datasheets and so on.

jbtronics:
> Regarding hardcoded LCSC supplier number import I agree, but will BatchInfoProvider interface allow to search by supplier numbers?

I would say that the batch search will just work like the single searches and search for a keyword. What the provider does with that is, is the choice of the provider (and will be limited by the API). Maybe one can pass an optional argument to demark what kind of information this is, to give the provider a hint. That could also be useful to have on the single search

> I haven't really tested with other than LCSC and Mouser as I don't regularly use others, but on them at least searching by MPN on batch input returns a very large number of results

Thats why I would implement it as on seperate interface. If it can be useful done for an provider (and somebody wants to implement it for an interface), then the batch methods will be used, otherwise not (and they do not exist on the provider then).

It would probably be the easiest if you could rewrite the controller to the more generic system of bulk info providers. I tried it yesterday and saw no easy way to do so, and would probably need more time to understand how the current code works...

jbtronics:
@barisgit Do you want to do the changes, or should i try it when I have time?

barisgit:
I will try to refactor, do I maybe create a new branch from v2.0 or will you then resolve merge conflicts after i refactor?

jbtronics:
This branch https://github.com/Part-DB/Part-DB-server/tree/feature/batch-info-provider-import
contains your changes rebased already on the v2 version (and most changes required already done), that should be easy to merge later. It also have some small improvements to your code, mostly related to the excel import/export.

I can also force push it into this PR if you want.

barisgit:
Yes that would be nice, as it is already opened.

jbtronics:
done.

barisgit:
All done, it should be better now, one thing i did was just refresh page (with js) after clicking a button in a bulk import job, maybe this code be done better. Also, why are github actions not running?

barisgit:
@jbtronics Can you take a quick look at structure to see if it is better so you don't have to wait for me if you find out there are additional things to do

jbtronics:
Its much better now. The controller is good now I think, extracting most of the logic into a service sounds good too.

Still, I do not really like the large complex arrays that get passed around (especially between different logical units). The performBulkSearch in the BulkInofProviderService returns a large complex array, which is basically impossible to understand without looking into the source code. I think returning a DTO would be helpful for making that easier to understand what it contains and more maintainable in the long run. For the internal passing around of data inside the service the array is okay, but then the arrays should have some type and or structure hints for parameter and return value, otherwise its very hard to understand what is done there... I have already started to add some typehints, but they are still missing in some places.

The field mapping could maybe also be extracted into a DTO, instead of an array, as it is used in many places. And the form could directly interact on these objects then.

Im not too sure about how useful the BulkSearchRequestDTO is, because it is only used in the one method signature, and is directly created before the method call and split up immediatley afterwards again. Here replacing the dto with named arguments in the performBulkSearch is probably doing the same.

I do not think its too useful to have BulkSearchResultDTO as child class of SearchResultDTO, as this needs to be updated everytime  a new field is updated. I think it is more useful to just encapsulate the SearchResultDTO there. But i am not sure how it is used later.
Also its name is at least a bit misleading, as it is not really the counterpart to BulkSearchRequestDTO (in the sense that is directly returned by the performBulkSearch), nor returned by the BatchInfoProviderInterface

barisgit:
@jbtronics Please tell me if you see any more bugs or areas for improvement, also i fixed a bug where "Add Mapping" wasn't working.

jbtronics:
I did some refactoring to remove the remains of the arrays everywhere. Also did some other improvements.

I have now merged the PR, thank you.

Maybe some notes: The structure how the search results are serialized to the DB have changed slightly to your version, so if you had db entries already, you should probably delete all import jobs.

In the `BulkInfoProviderImportController::researchAllParts()` method, you used the entitymanager clear method to save a bit of memory. This caused problems in my tests (the flush complained about a missing user entity), so i commented the clear out.

For the future, it would probably be cool to make the whole data retrieval thing which can take some time truely asynchronous from the web tasks. This should be not so difficult with the symfony/messenger component, but so far Part-DB does not use it yet.

PR #983, opened 2025-08-03
Ability to upload schematic BOMs to projects, not only PCB BOMs

barisgit:
As #907 already outlines, currently only ability to import BOMs from KiCAD PCBNew is allowed. With this PR, there is now functionality to import BOMs from schematic as well as any other arbitrary CSV, thanks to a mapping step that maps input columns to output columns.

This is how it works:
1. User selects "Import BOM" from project view where they opt to use the second option - Schematic BOM (supports any arbitrary CSV)
Screenshot 2025-08-03 at 18 41 56

2. The file is parsed for columns, which are then presented to be mapped to data in PartDB (includes smart mapping suggestion, where users shouldn't need to change much)
Screenshot 2025-08-03 at 18 41 39

3. There is also ability to select priority in case there are multiple columns that would map to same column
Screenshot 2025-08-03 at 18 41 47

4. Upon submitting, server tries to connect parts to existing parts in the database, for others users can choose as is done with PCB BOMs as well
Screenshot 2025-08-03 at 18 43 31

Note: The system also validates CSVs and provides error messages in cases of invalid data
Screenshot 2025-08-03 at 18 41 18

Lines also get consolidated in case there are same components on different lines. Eg.:
- Line1: C1, C3, C5; Qty 3; MPN 123456
- Line 6: C6, C12, C13, C24; Qty 4; MPN 123456

would get saved as 7 parts with MPN 123456 of which designators are C1, C3, C5, C6, C12, C13 and C24.

### DIFF
--- a/docs/usage/bom_import.md
+++ b/docs/usage/bom_import.md
@@ -34,3 +34,12 @@ select the BOM file you want to import and some options for the import process:
   has a different format and does not work with this type.
   You can generate this BOM file by going to "File" -> "Fabrication Outputs" -> "Bill of Materials" in Pcbnew and save
   the file to your desired location.
+* **KiCAD Schematic BOM (CSV file)**: A CSV file of the Bill of Material (BOM) generated
+  by [KiCAD Eeschema](https://www.kicad.org/).
+  You can generate this BOM file by going to "Tools" -> "Generate Bill of Materials" in Eeschema and save the file to your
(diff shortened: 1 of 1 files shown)


barisgit:
I am sorry about these shenanigans, the branch is back to where it should be.

jbtronics:
Thanks, that great.

Merged

PR #984, opened 2025-08-03
Make image size in table values settable by env

barisgit:
See #623.

I couldn't continue on that branch, as it is too old.


jbtronics:
As we now have the new settings system, I rewrote this in a separate feature. It offers the same env variables, so that it should be compatible.

Therefore this PR is obsolete.

Issue #577, opened 2024-03-28
Collect parts for a later action like printing labels


frank-f:
**Is your feature request related to a problem? Please describe.**
I am currently going through my cabinets and drawers, collecting all parts I can find and adding them to Part-DB. To have them better organized afterwards I am also printing a label for each individual part. For that I am using A4 sheets with pre-cut stickers. That means, the handling is very inefficient for me if I print every label individually. It is best to wait until I have enough parts to fill a complete sheet and print that in one go.

**Describe the solution you'd like**
Perfect would be a "Label printed" flag for part lots (+ parts?) (similar to "needs review" for parts), that is unset by default. I would need a filter option above part lists and could then jump to the pre-filled label generation page via the action menu. Also the label generation page could include a "set label printed flag" checkbox that would allow me to easily flip the flag for exactly those items that I just printed.

**Describe alternatives you've considered**
An alternative, that would probably require less structural changes and is even more flexible, would be some way to mass-add and remove tags from parts (again, similar to "needs review" in the action menu) and use those to identify unprinted part lots (= parts). I'd add the tag when I create a part and then remove the tag once a full sheet was printed. This does not work well for parts consisting of multiple lots however.

**Additional context**
Another thing I thought of while printing the first labels was, it'd be really handy if I could specify a number of empty labels to be added before the first real one. That would make it a bit easier to print on sheets that have already been used partially.

PR #995, opened 2025-08-14
Implement N-up label printing

d-buchmann:
Make label printing more flexible by allowing to put multiple labels onto a single PDF page in a grid ("N-up printing").
This way you can use e.g. A4 sheets with pre-cut labels and print them in your household inkjet/laser printer - no need for a special label printer.

The grid of the sheet is set by `xcount` and `ycount` which I added to the form (but not yet to the database).
You can also set a `skipcount` which skips the given number of labels in case you have partly printed sheets from previous prints (feature as described/requested in #577).
The size and position tweaking can be done in the additional css to fit any label sheet type.

As far as I have tested, it doesn't break the original behavior using the default values `xcount=ycount=1, skipcount=0`.
Currently translations and migrations are yet to be added.


z3ugma:
I would benefit a lot from this PR getting merged. What do you think @jbtronics  ?

src/Entity/LabelSystem/LabelOptions.php
```diff
@@ -65,6 +65,12 @@ class LabelOptions
     #[Groups(["extended", "full", "import"])]
     protected float $height = 30.0;
 
+    // TODO these have to go into the DB to be persisted.
```
z3ugma: @d-buchmann do these TODOs need to be resolved before merge, or are they optional enhancements for later

d-buchmann: It does work as-is, but the grid parameters are not stored in the label profile yet, which is why I'd hesitate.
`skipcount` honestly has no significance in this context, it's an ephemeral value just like the other inputs when generating a label.

PR #997, opened 2025-08-14
UX: Always show the select panel

d-buchmann:
Make the select panel always visible because several user reports suggest that they are unaware of the feature.
Also solves the minor issue that selecting/deselecting a single row changes the position of the table which makes it prone to misclicks.

Note: Translations have been adjusted in place, changes should be reflected in crowdin in case this will be merged.

### DIFF
--- a/assets/controllers/elements/datatables/parts_controller.js
+++ b/assets/controllers/elements/datatables/parts_controller.js
@@ -42,11 +42,11 @@ export default class extends DatatablesController {
 
         const selectPanel = this.selectPanelTarget;
 
-        //Hide/Unhide panel with the selection tools
+        //Toggle action button based on selection
         if (count > 0) {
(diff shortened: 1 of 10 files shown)


jbtronics:
Im not a huge fan of this, as without selection this bar is just useless.
I have now implemented an alternative, which i hope should have the same advantages. The action bar is now sticky floating over the table. So you should notice it quite easily when clicking a checkbox.

image

This also helps, that you do not have to scroll back to the top, after doing the selection.

d-buchmann:
This looks very nice.
I'm fine with closing this PR, but I'd appreciate if you consider cherry-picking the translation update [7993228](https://github.com/Part-DB/Part-DB-server/pull/997/commits/799322846912d12405581f32af9751f368a854e7) ;)

jbtronics:
done

PR #1021, opened 2025-09-04
Add import permission for label profiles

### LINKED ISSUE #1020: Label profiles can not be imported
**Describe the bug**
Label profiles can not be imported.
I suspect this is because there is no permission for the import of label profiles.

**To Reproduce**
1. Go to Label profiles
2. Click on Import/Export tab
3. Import is greyed out despite admin access with full permissions granted

**Expected behavior**
I'd like to be able to import label profiles.
Export works as expected.

**Server Side**
- Part-DB Version: 1.17.1 resp. 2.0.2
- PHP Version: 8.2 resp. 8.4
- Database Server MariaDB 10.3 resp. MySQL 8

**Additional info**
Working with Part-DB mostly as admin, I found it odd that fields were disabled.
A tooltip describing why something can't be done (e.g. lacking permissions) would be nice.

d-buchmann:
Fixes #1020.

In the long run, I think the permissions for labels and label profiles should be separated.

### DIFF
--- a/config/permissions.yaml
+++ b/config/permissions.yaml
@@ -359,6 +359,10 @@ perms: # Here comes a list with all Permission names (they have a perm_[name] co
         label: "perm.revert_elements"
         alsoSet: ['read_profiles', 'edit_profiles', 'create_profiles', 'delete_profiles']
         apiTokenRole: ROLE_API_EDIT
+      import:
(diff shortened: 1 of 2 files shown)


jbtronics:
Thanks

PR #1027, opened 2025-09-07
Add INITIAL_ADMIN_API_KEY environment variable support

rsp2k:
- Add configuration parameter for initial admin API key
- Implement getInitialAdminApiToken() method in AbstractMultiPlatformMigration
- Create migration to automatically generate admin API token on initial setup

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,6 @@ yarn-error.log
 ###> phpstan/phpstan ###
 phpstan.neon
 ###< phpstan/phpstan ###
+
+# Claude Code project documentation (local development only)
+CLAUDE.md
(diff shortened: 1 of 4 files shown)


jbtronics:
What is your use case for this?
I don't have much of a problem with such a feature, but I never thought that having an API key from the start would be useful?

And maybe a console command to programmatically create API keys would be more useful than doing that in the database migration ?

rsp2k:
I'm using it to bootstrap servers while testing. Thank you so much for this absolutely wonderful project!  Hats off to you sir!

jbtronics:
What would you say about restricting the lifetime of the API token to a week (or even just a day)?
That should be enough for testing and bootstrapping applications and should improve security, as you do not risk unauthorized access, if you somehow forget this token or its secret get leaked.

Also I would not put the env into the docker-env templates. That is an feature for advanced usecases, and if you put it into the docker-compose file, then it will probably stay there forever, risking it to be leaked somehow...

PR #1051, opened 2025-09-22
Feature Assemblies / Custom part state / Part IPN suggest with category prefixes

webdevinition:
In relation to the discussion opened at https://github.com/Part-DB/Part-DB-server/discussions/1041, I would like to submit this pull request.

The described features have already been merged into an "all-feature" branch, which should make it easier to merge the changes. The branch mentioned has been rebased to version 2.1.2 of Part-DB.
I'm very much looking forward to a future integration of this!

Thank you in advance!

PR #1053, opened 2025-09-24
Feature custom part status

webdevinition:
Hello @jbtronics,

In relation to https://github.com/Part-DB/Part-DB-server/pull/1051 the feature for custom part states.
Also mentioned in the discussion under https://github.com/Part-DB/Part-DB-server/discussions/1041, section 3.

I am grateful for integration!

Best regards,
Marcel


jbtronics:
My preliminary remarks from a quick look at the PR:

* What is the advantage of having the custom states as its own entity over just putting that into a tag? (Or just an additional text field in the part)?
* The two migrations should probably be merged. I dont see much reason to have it split in two.
* You should add tests, to ensure the admin pages are available and everything behaves like intended
* Something has broken the existing tests.

webdevinition:
@jbtronics, please excuse me for not getting back to you on this matter yet.
It has not been forgotten, and feedback will follow in any case.

webdevinition:
Hello @jbtronics, thank you for your patience.

Regarding your comments:

> What is the advantage of having the custom states as its own entity over just putting that into a tag? (Or just an additional text field in the part)?

In everyday use, it has proven useful to be able to use functions similar to those of the other entities, such as uploading descriptive attachments or hierarchically storing desired status types. The possibility of storing notes is also not to be overlooked. This allows the administrator to specify at the time of creation exactly when the status should be used in everyday life and why it was ultimately created.

> The two migrations should probably be merged. I dont see much reason to have it split in two.

That's done.

> You should add tests, to ensure the admin pages are available and everything behaves like intended

Now there are also corresponding tests which I have added analogously to other entities.

> Something has broken the existing tests.

I recently made an adjustment, which may have caused the problem.

Thanks for everything so far! I've pushed the additional adjustments.

jbtronics:
I wonder if it would be useful that a part can have more than one custom state. Maybe in combination with a simple constraint that ensures that only one status out of a single parent type is selected. That way the system is a bit more universal, and would allow for different kind of states...

I have also fixed the test issues and some translation inconsistency (for every other entity the plural is used in the admin page title and tree node. I just changed it for english as in german this would lead to some weird words). However i cannot push the changes, as it tells me "Permission denied".

Besides these points, i would say the PR is ready for merge.

webdevinition:
@jbtronics, thank you very much for your efforts and preparation! I've checked the "Allow maintainer changes" box in the pull request.

We haven't had to set multiple statuses in practice yet, but there might be other Part-DB users who would welcome something like that.

I am very pleased about the trusting cooperation, which I enjoy, and your comments and improvements :)

Best regards,
Marcel

blue3121:
Can you add, for completeness, multiple touch on "custom part status" for selected parts ?
Thanks !
(Wonderful work, good job !!!)

PR #1054, opened 2025-09-25
Feature part IPN suggest with category prefixes

webdevinition:
Hello @jbtronics,

In relation to https://github.com/Part-DB/Part-DB-server/pull/1051 the feature for part IPN suggest with category prefixes.
Also mentioned in the discussion under https://github.com/Part-DB/Part-DB-server/discussions/1041, section 2.

I am grateful for integration!

Best regards,
Marcel

### DIFF
--- a/.env
+++ b/.env
@@ -50,6 +50,8 @@ EMAIL_SENDER_EMAIL=noreply@partdb.changeme
 EMAIL_SENDER_NAME="Part-DB Mailer"
 # Set this to 1 to allow reset of a password per email
 ALLOW_EMAIL_PW_RESET=0
+# Set this to 0 to allow to enter already available IPN. In this case a unique increment is appended to the user input.
+ENFORCE_UNIQUE_IPN=1
 
 ###################################################################################
 # Error pages settings
@@ -115,6 +117,9 @@ REDIRECT_TO_HTTPS=0
 
 # Set this to zero, if you want to disable the year 2038 bug check on 32-bit systems (it will cause errors with current 32-bit PHP versions)
 DISABLE_YEAR2038_BUG_CHECK=0
+
+# Define the number of digits used for the incremental numbering of parts in the IPN (Internal Part Number) autocomplete system.
+AUTOCOMPLETE_PART_DIGITS=4
 
 # Set the trusted IPs here, when using an reverse proxy
 #TRUSTED_PROXIES=127.0.0.0/8,::1,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16
--- a/.env
+++ b/.env
@@ -50,8 +50,6 @@ EMAIL_SENDER_EMAIL=noreply@partdb.changeme
 EMAIL_SENDER_NAME="Part-DB Mailer"
 # Set this to 1 to allow reset of a password per email
 ALLOW_EMAIL_PW_RESET=0
-# Set this to 0 to allow to enter already available IPN. In this case a unique increment is appended to the user input.
-ENFORCE_UNIQUE_IPN=1
 
 ###################################################################################
 # Error pages settings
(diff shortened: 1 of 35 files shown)


jbtronics:
Some preliminary remarks:

* Why do you drop the unique constraint on the IPN? I think it is important that the IPNs are truly unique, otherwise they are not useable as identifiers. And as far as I understand the `IPN_ENABLE_UNIQUE_CHECK` options documentation, an additional suffix is added anyway.
* Im not sure if the name "enableUniqueCheck" is maybe a bit misleading. Maybe something like "automatically attach suffix or something" might be better.
* It might be a good addition to have the option to define a global regular expression, that IPNs have to fullfill. that way you can enforce certain formats.
* The repo method, should not require get a base64 decoded string. That decoding should happen in the controller that handles the request.
* Is it really useful to generate the IPN based on the part description? Does this gives much advantages about just using the part category?
* There should be test ensuring the correct behavior of the IPN suggestions and the autocomplete suggestion endpoint.

jbtronics:
Also you should fix the complains of phpstan.

webdevinition:
@jbtronics, please accept my apologies for not having contacted you about this matter yet.
I haven't forgotten about it, and I will definitely get back to you.

I have also worked on some of the points you mentioned so that I can get as close as possible to what you requested.

jbtronics:
No worries, take your time ;)
I will be in vacation next week and won't be able to do anything then.

webdevinition:
Hello @jbtronics, thank you for your understanding!
I was also on vacation last week. Hope you had a good time!

Regarding your comments:

> Why do you drop the unique constraint on the IPN? I think it is important that the IPNs are truly unique, otherwise they are not useable as identifiers. And as far as I understand the IPN_ENABLE_UNIQUE_CHECK options documentation, an additional suffix is added anyway.

This is, of course, completely correct. I've reverted this and renamed the constant (see below).

>  Im not sure if the name "enableUniqueCheck" is maybe a bit misleading. Maybe something like "automatically attach suffix or something" might be better.

Thanks for the tip. I've renamed it to "autoAppendSuffix/IPN_AUTO_APPEND_SUFFIX" Hopefully, that makes it clearer.

> It might be a good addition to have the option to define a global regular expression, that IPNs have to fullfill. that way you can enforce certain formats.

Thank you, that's a good point. I've added IPN_SUGGEST_REGEX and IPN_SUGGEST_REGEX_HELP for help text that the Regex administrator can add for their Part-DB users, which will then also be displayed in the Part configuration (see below).

> The repo method, should not require get a base64 decoded string. That decoding should happen in the controller that handles the request.

Thanks, I changed that.

> Is it really useful to generate the IPN based on the part description? Does this gives much advantages about just using the part category?

I have a use case that occurs frequently in everyday life. For example, there are often repetitive description texts for multiple components, and in these cases, it is helpful to have a search/suggest with the next possible part increment after inserting an existing description text.
Since not everyone wants to use this, it is now disabled by default and must first be enabled via SystemSettings  / $useDuplicateDescription / IPN_USE_DUPLICATE_DESCRIPTION. If disabled, no description is sent to the suggest endpoint. I hope this is okay for you.

> There should be test ensuring the correct behavior of the IPN suggestions and the autocomplete suggestion endpoint.

I have implemented a test for this.

Attached is a screenshot of the implemented system settings:

2025-10-13-Part IPN Suggest - Server settings

Thanks for everything so far! I've pushed the adjustments.

jbtronics:
I have improved/fixed some smaller things, and merged the PR. Thanks for your efforts. Great feature.

Maybe some ideas for further improvements:
* The ability to customize the separator between category prefix and prefix levels
* Allow to set a global prefix (as fallback if a category do not have one or in general)

PR #1151, opened 2025-12-11
Buerklin provider

mkne:
This pull request implements the Buerklin API and as Information Provider that also implements the BatchInfoProviderInterface.
Bürklin also offers an API. (Apply here: https://www.buerklin.com/de/services/eprocurement/)
Authorization works via Bearer Token that is requested via a POST-action to https://www.buerklin.com/authorizationserver/oauth/token/ and the following parameters:
client_id:cws-client
client_secret:{{PROVIDER_BUERKLIN_SECRET}}
grant_type:password
username:{{PROVIDER_BUERKLIN_USERNAME}}
password:{{PROVIDER_BUERKLIN_PASSWORD}}

There are only two GET-actions available:

Product search (example with the part number 66S447): https://www.buerklin.com/buerklinws/v2/buerklin/products/66S447/?curr=EUR&language=de
Free text search (example with the query string 'Resistor'): https://www.buerklin.com/buerklinws/v2/buerklin/products/search/?curr=EUR&language=en&pageSize=50&currentPage=0&query=Resistor&sort=relevance
I would set the parameters curr and language via PROVIDER_BUERKLIN_CURRENCY and PROVIDER_BUERKLIN_LANGUAGE.

Bürklin supplies a postman collection which I used to test the API and reverse engineer.
For reference: https://github.com/Part-DB/Part-DB-server/discussions/868

### DIFF
--- a/src/Services/InfoProviderSystem/Providers/BuerklinProvider.php
+++ b/src/Services/InfoProviderSystem/Providers/BuerklinProvider.php
@@ -58,6 +58,7 @@ class BuerklinProvider implements BatchInfoProviderInterface
 
     /**
      * Gets the latest OAuth token for the Buerklin API, or creates a new one if none is available
+     * TODO: Rework this to use the OAuth token manager system in the database...
      * @return string
      */
     private function getToken(): string


src/Services/InfoProviderSystem/Providers/BuerklinProvider.php
```diff
@@ -0,0 +1,633 @@
...
+    }
+
+
+    public function getProviderInfo(): array
+    {
+        return [
+            'name' => 'Buerklin',
+            'description' => 'This provider uses the Buerklin API to search for parts.',
+            'url' => 'https://www.buerklin.com/',
+            'oauth_app_name' => self::OAUTH_APP_NAME,
```
d-buchmann: If the auth requires no user interaction (ever), you can delete this line.

src/Services/InfoProviderSystem/Providers/BuerklinProvider.php
```diff
@@ -0,0 +1,633 @@
...
+
+
+    /**
+     * Returns a valid Buerklin product short URL from product code
+     * @param  string  $product_code
+     * @return string
+     */
+    private function getProductShortURL(string $product_code): string
+    {
+        return 'https://www.buerklin.com/de/p/' . $product_code . '/';
```
d-buchmann: I think this could use a sanity check of the product code (and localization).
We shouldn't expect the remote API to always behave well :)

d-buchmann (changes requested):

src/Services/InfoProviderSystem/Providers/PollinProvider.php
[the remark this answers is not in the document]
mkne: Reverted this commit.

src/Services/InfoProviderSystem/Providers/BuerklinProvider.php
mkne: I added the localization but I didn't add a regex to check for the format of the product_code because it is not guaranteed that Buerklin will change them in the future. We have a lot of Buerklin parts in our lists where the order number is just no longer existing.

src/Services/InfoProviderSystem/Providers/BuerklinProvider.php
```diff
@@ -0,0 +1,633 @@
...
+    ) {
+
+    }
+
+    /**
+     * Gets the latest OAuth token for the Buerklin API, or creates a new one if none is available
+     * @return string
+     */
+    private function getToken(): string
+    {
```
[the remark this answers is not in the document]
mkne: I will test this when I have access to my test server. Thanks for pointing in the right direction. But if it doesn't work I would stay with my current getToken because it works.

src/Services/InfoProviderSystem/Providers/BuerklinProvider.php
mkne: I wasn't sure if a login like with the Digikey Provider will be needed.

jbtronics:
Thanks.
I did not tested if it actually works, because i have no API token for buerklin, but i assume you did that and it is working fine.

I am also not a huge fan of this OAuth token mechanism, but for now this is fine, and without an api token im not able to find a better solution for this.

PR #1154, opened 2025-12-14
fix(digikey): prevent crashing when no media is found for part

hrueger:
Hi,
this fixes the following crash:
grafik


jbtronics:
I implemented this in a slighlty different alternative version, where only on a status code 404 an empty result returned. So that if any other kind of error occurs, the error is not silenced and can be handled somewhere else.

So this PR is obsolete

hrueger:
Funny, that is exactly what I did originally, in 2a21a703880842e526c8f8fc29c681ae41405464

Happy that this is fixed now though, so thanks!

Issue #1166, opened 2025-12-16
Provider capabilities should be shown in the same order


mkne:
This is mainly an pseudo bug report to be able to merge a pull-request for the Buerklin Info Provider.
To make the list of info provider a bit easier to compare. I moved the footprint to the back of the list.
<img width="876" height="595" alt="Image" src="https://github.com/user-attachments/assets/2c3d8154-f5ad-49a2-9b75-227635b98c2e" />

<img width="918" height="188" alt="Image" src="https://github.com/user-attachments/assets/8a2941a4-fea0-4c3e-8f5a-b9a630149aae" />

jbtronics:
I added an dynamical sorting that brings it into a fixed order

PR #1167, opened 2025-12-17
Fix Provider capabilities should be shown in the same order #1166

mkne:
Provide the ProviderCapabilities in the following order:
ProviderCapabilities::BASIC,
ProviderCapabilities::PICTURE,
ProviderCapabilities::DATASHEET,
ProviderCapabilities::PRICE,
ProviderCapabilities::FOOTPRINT,

Footprint seems to be the least common, so put it at the end of the list.


jbtronics:
I added an dynamical sorting that brings it into a fixed order, so this does not need to be changed for every info provider by hand and makes it easier if something is added that could change the order.
So this PR is obsolete

Issue #1161, opened 2025-12-15
Show Part ID in Project BOM overview


fsbrc:
Is your feature request related to a problem? Please describe.
- In the Project BOM overview it is currently not possible to see the actual Part ID of the referenced part. Only the IPN and BOM entry / project-related ID are possible columns

Describe the solution you'd like
- Add an additional column to the Project BOM overview that displays the Part ID of the linked Part entity.
- The column should show the real Part database ID (part.getId()), not the Project BOM entry ID.

Additional context
For someone who is able to code this is probably a really easy feature request.
Likely implementation location: src/DataTables/ProjectBomEntriesDataTable.php
Possible implementation approach:
Add a TextColumn that renders the Part ID via the ProjectBOMEntry, e.g. by accessing getPart()->getId().
This feature would significantly improve usability when working with large BOMs and when referencing parts across projects.

`            ->add('part_id', TextColumn::class, [
                'label' => 'Part-ID',
                'render' => function ($value, ProjectBOMEntry $context) {
                    return $context->getPart()?->getId() ?? '';
                },
            ])`

It would be really great if one could search for the Part-ID via the search bar as well just like the IPN

jbtronics:
done

PR #1168, opened 2025-12-18
added feature of part-id in project bom view

fsbrc:
Adds a new visible column part-id to the Project BOM overview

- The column displays the real Part database ID (Part::getId()), not the Project BOM entry ID
- Column is sortable via part.id

For BOM entries without a linked Part, the column remains empty
Trying to add https://github.com/Part-DB/Part-DB-server/issues/1161


jbtronics:
thanks

PR #1169, opened 2025-12-22
added part.id to search options and to bom

kernchen-brc:
- [x] Added "ID" to the search options.
- [x] Added a "Part ID" column to the project BOM, using the actual part.id instead of the part.table.id.  
      This implements the same improvement as fsbrc’s pull request but uses the translator instead of a static string.


jbtronics:
I am not really sure if I like the LIKE behavior for the id search. If you type in "56" it will also show you the parts with ID "561", "156", "5600", and so on.
In combination with enabling that field by default (and hiding the ID column) this will probably lead to a lot of undesired search results, when searching for a number.

Also, I am not sure if this actually works on all database platforms (like postgres), as it's applying a LIKE operator to an integer column somewhere.

jbtronics:
I also merged PR #1168 for the part id in BOM tables.

PR #1171, opened 2025-12-28
Fix #1165: Label generator PDF viewer could be a few px longer

### LINKED ISSUE #1165: Label generator PDF viewer could be a few px longer
Going to put that as a feature request I guess so, the label generator does have a pdf viewer, I guess the workflow is to use the download button, open the pdf, then print it, however on firefox it does shows the bar with zoom/print/download etc. but on vivaldi (and other chrome-ium based ones I guess) it doesn't ... unless it's made a bit more longer with the bottom right arrow and tada, the header shows!

I think having it a bit more longer by default so the header shows would be better, I always have to make it longer then use the print feature directly to not have 400 label(432).pdf on my download folder :D

firefox:

Image

vivaldi default:

Image

made it a bit longer:

Image

mkne:
Increase default height of the PDF preview container from 250px to 280px and so Chromium-based browsers display the PDF toolbar by default. Fixes #1165.

### DIFF
--- a/templates/label_system/dialog.html.twig
+++ b/templates/label_system/dialog.html.twig
@@ -135,8 +135,8 @@
 
 {% block additional_content %}
     {% if pdf_data %}
-        <div class="card mt-2 p-1 border-secondary" style="resize: vertical; overflow: scroll; height: 250px">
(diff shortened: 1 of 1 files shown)


jbtronics:
Thanks

PR #1184, opened 2026-01-09
Added Exact ID Search Option (Disabled by Default)

kernchen-brc:
I reworked my latest pull request. Now it introduced an ID search option that is disabled by default. ID search now uses exact match (=) instead of LIKE.

The filter applies only when:
- The ID search option is enabled and
- the keyword is numeric.

 Updated apply() in PartSearchFilter.php to implement this logic while preserving previous functionality for other fields.


jbtronics:
Looks good. Thank you.

PR #1187, opened 2026-01-12
Update OEMSecretsProvider.php

### LINKED ISSUE #1178: 500 error when creating a part via OEMSecrets
Part-DB
Version: 2.4.0

I can reproduce a 500 error when creating a part via OEMSecrets. Container logs show a PHP error:
"Object of class ... OEMSecretsProvider could not be converted to string" (OEMSecretsProvider.php:683).

From the same container network namespace, the exact OEMSecrets request URL taken from the Part-DB logs returns HTTP 200 and valid JSON:

1) Extract last OEMSecrets URL from logs (sanitized):
URL="$(sudo docker logs --tail 5000 partdb 2>&1 | grep -oE 'https://oemsecretsapi\.com[^\" ]+' | tail -n 1)"

2) Verify status code:
sudo docker run --rm --network container:partdb curlimages/curl:8.5.0 \
  -sS -I -o /dev/null -w "HTTP %{http_code}\n" "$URL"
--> HTTP 200

3) Verify response is JSON:
sudo docker run --rm --network container:partdb curlimages/curl:8.5.0 \
  -sS "$URL" | head -c 400
--> starts with: {"version":"3.0","status":"http 200 OK",...}

So the OEMSecrets endpoint is reachable and returns JSON; the 500 seems to be caused by a bug in Part-DB while handling the OEMSecrets response (likely an exception message / string conversion issue at OEMSecretsProvider.php:683).

d-buchmann:
Most probably only a typo.
Should fix #1178, but I didn't test it honestly.

### DIFF
--- a/src/Services/InfoProviderSystem/Providers/OEMSecretsProvider.php
+++ b/src/Services/InfoProviderSystem/Providers/OEMSecretsProvider.php
@@ -680,7 +680,7 @@ class OEMSecretsProvider implements InfoProviderInterface
         if (is_array($prices)) {
             // Step 1: Check if prices exist in the preferred currency
             if (isset($prices[$this->settings->currency]) && is_array($prices[$this->settings->currency])) {
-                $priceDetails = $prices[$this->$this->settings->currency];
+                $priceDetails = $prices[$this->settings->currency];
                 foreach ($priceDetails as $priceDetail) {
                     if (
                         is_array($priceDetail) &&


jbtronics:
Yeah, seems like there was some typo, when switching to the settings system.
I would have expected that phpstan, catches such things, but it seems it does not.
Unfortunatley its also hard to actually test the providers (as that would make requests to the providers), so our continous testing does not really work either.

Thanks for fixing.

Issue #1095, opened 2025-11-04
oo many BOM items; error message: "The CSRF token is invalid. Please try to resubmit the form."


pipeblue0608:
When uploading a BOM with an extremely large number of items, it cannot be edited and an error is prompted.

**To Reproduce**
1.the BOM in the uploaded attachment

[PBZU19-BOM_sorted_merged_converted.csv](https://github.com/user-attachments/files/23325929/PBZU19-BOM_sorted_merged_converted.csv)

2. edit
3. See error: The CSRF token is invalid. Please try to resubmit the form

**Server Side**
- Part-DB Version: [e.g. Part-DB 2.2.1 (for dev versions commit hash is useful]
- PHP Version: [e.g. PHP 8.4.]
- Database Server [e.g. sqlite]

d-buchmann:
Duplicate of #911

pipeblue0608:
For users with a large number of BOM types, this bug is fatal.

bernde78:
Hi,
I have the same issue with PartDB 2.13.4! 
We have some bigger projects with more then 141 entries in the BOM. When I import such a BOM and try to select the part in the BOM and save it, I will show an error Invalid CSRF-Token .

When I delete some parts from the BOM so that there are only 141 entries in the BOM, it works without the error and I can select the parts for all BOM entries and save it without any problem.

Is there a fix?

Kind regards,
Bernd


**UPDATE:
Increasing the limit max_input_vars to 2000 in php.ini fixed the problem for me!**

d-buchmann:
Nice that you found a reasonable workaround.
That being said, the first reporter of this bug recently issued PR #1488 to fix it.

jbtronics:
Should be fixed with the merge of #1488 and recommendation of increasing max_input_vars

Issue #911, opened 2025-04-15
Problem Adding and Editing Parts in Large Projects


T0b121:
**Describe the bug**
When adding parts to a project in Part-DB, an error occurs after a certain number of entries. The issue manifests in the following scenarios:

1. **Adding Parts:** After a certain number of parts (tested in increments of 100 from 100 to 1000), no further parts can be added.
2. **Editing Projects:** After a certain number of parts (e.g., 200), the project can no longer be moved or edited. The error "CSRF token is invalid" occurs.
3. **Moving Projects:** When moving projects to another folder, the part names are deleted, and the error "CSRF token is invalid" occurs.
4. **Importing Parts:** When importing parts into projects with 600 or more items, an error occurs. The projects can be viewed in the project tab but cannot be edited via the edit tab.
5. **Demo Website:** The same error occurs on the demo website (https://demo.part-db.de/en/project/187/edit#bom) when editing a project with a large number of parts.

**To Reproduce**
**Problem with Edit of a Project**
1. Create a new project in Part-DB.
2. Import KiCAD BOM with many parts (attached are BOM with a multiple of 100: 100, 200, 300, ..., 1000).
3. Attempt to add more parts, move Project ore edit it over all when it has 200 or more parts.
**Edit Page is not more able to open**
1. Import parts into a project with 600 or more items.

**Expected behavior**
- Parts should be added without issues, regardless of the number.
- Projects should be editable and movable, regardless of the number of parts.
- Parts import should work regardless of the number of items in the project.
- The demo website should function correctly when editing projects with a large number of parts.

**Screenshots**
- Image 1_Project with 200 Parts befor move.png ![Image](https://github.com/user-attachments/assets/c88ba5ab-63f2-4c31-802d-813bd5864e82)
- Image 2_Project with 200 Parts after move.png ![Image](https://github.com/user-attachments/assets/7f4ea987-3a2c-4f0c-ae82-1d66c6ac17af)
- Image 3_Open Edit from Project with 600 Parts.png ![Image](https://github.com/user-attachments/assets/e0bda6f3-4ec7-4a8a-8d12-c34f527f9a83)

**Server Side**
- Docker 20.10.24+dfsg1
     - Part-DB Version: [1.17.0]
     - Database Server [MariaDB Community 11.7.2]

**Desktop:**
 - OS: [Windows 11]
 - Browser [Edge]
 - Version [135.0.3179.54 64-Bit]

**Additional context**
PartDB config
[...]

PR #1191, opened 2026-01-15
Change barcode scan form to use requestSubmit() to fix CSRF token error

swdee:
On the Label Scanner page I have developed scanning of LCSC barcodes.  During testing the redirect to the Part page was not working and I was getting CSRF token errors and the request failing with a HTTP 422 error.

I noticed that the CSRF Token value being sent on the form submission had the placeholder text  `scan_dialog[_token]
csrf-token `, so the token itself was not being generated and submitted.

In the assets/controllers/csrf_protection_controller.js  file it specifically states;
```
// Generate and double-submit a CSRF token in a form field and a cookie, as defined by Symfony's SameOriginCsrfTokenManager
// Use `form.requestSubmit()` to ensure that the submit event is triggered. Using `form.submit()` will not trigger the event
// and thus this event-listener will not be executed.
```

Sure enough the Label Scanner form submission was using `form.submit()` and not the required `form.requestSubmit()`.    This PR patches that bug so CSRF token is now correctly generated on form submission.

I further note that this bug maybe present else where in the code as a grep shows;
```
$ rg -n "\.submit\(\)|requestSubmit\(" -S assets/controllers

assets/controllers/csrf_protection_controller.js
5:// Use `form.requestSubmit()` to ensure that the submit event is triggered. Using `form.submit()` will not trigger the event

assets/controllers/pages/barcode_scan_controller.js
73:        document.getElementById('scan_dialog_form').submit();

assets/controllers/helpers/form_cleanup_controller.js
65:        form.submit();

assets/controllers/elements/part_search_controller.js
122:                input.form.requestSubmit();

assets/controllers/elements/datatables/parts_controller.js
143:        //We need to do this that way, as we need the submitter info, just calling form.submit() would not work
```
The other possible bug is in the `assets/controllers/helpers/form_cleanup_controller.js` file, but as I am not sure which part of the code this effects, I have not proposed a change here.  However that could be the failure for other CSRF issues in #911 and #1095

### DIFF
--- a/assets/controllers/pages/barcode_scan_controller.js
+++ b/assets/controllers/pages/barcode_scan_controller.js
@@ -70,6 +70,6 @@ export default class extends Controller {
         //Put our decoded Text into the input box
         document.getElementById('scan_dialog_input').value = decodedText;
         //Submit form
-        document.getElementById('scan_dialog_form').submit();
+        document.getElementById('scan_dialog_form').requestSubmit();
     }
 }


jbtronics:
Thanks.

I do not think that these issues are related to this however. For very large forms, you will run into the variable limit of PHP and then the csfr field is not submitted properly.

PR #1192, opened 2026-01-15
fix(KiCad API): :bug: new categories excluded from simulation

lukas-runge:
Hi @jbtronics,

first of all - thank you for developing and maintaining this incredible piece of software. 🙏

When initially adding all my parts to `Part-DB` with the defaults provided, I just discovered that categories get created with `excluded_from_sim` set to `true`. When also creating all parts with the default of `null` (which means `inherit` from category) then all parts are per default excluded from simulation when adding them to KiCad via the API endpoint. This means that in older versions of KiCad (pre v9) newly placed parts are initially shown crossed out

Bildschirmfoto 2026-01-15 um 19 27 48

and in newer versions (v9 and newer) they get outlined and shown with an additional tag on the bottom right side

Bildschirmfoto 2026-01-15 um 19 29 26

indicating that they are excluded from simulation.

This is rather annoying as changing this now requires manual adjustment of that setting for every single category in existence. This could only be made easier for existing users by providing some kind of batch editing feature which as far as I know is not a thing.

To at least mitigate this inconvenience for new users and new categories that existing users create, this PR proposes to set `excluded_from_sim` to `null` for new categories and default to `false` in the helper function if the property is not set.

Best regards,
@lukas-runge

### DIFF
--- a/src/Entity/EDA/EDACategoryInfo.php
+++ b/src/Entity/EDA/EDACategoryInfo.php
@@ -58,7 +58,7 @@ class EDACategoryInfo
     /** @var bool|null If this is set to true, then this part will be excluded in the simulation */
     #[Column(type: Types::BOOLEAN, nullable: true)]
     #[Groups(['full', 'category:read', 'category:write', 'import'])]
-    private ?bool $exclude_from_sim = true;
+    private ?bool $exclude_from_sim = null;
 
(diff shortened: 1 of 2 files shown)


jbtronics:
I would not really call it a bug, but sounds reasonable to change the default behavior here.
Thanks.

PR #1208, opened 2026-01-25
Support dynamic supplier SPNs in BOM import comments

MayNiklas:
I split this PR into 4 commits to be easier to review step by step.
Commit 1+2 fix a bug.
Commit 3+4 extends functionality around the bug fixed  and therefore should be treated as being optional (if you don't like it).

Issue:
The field mapping system uses the pattern `$supplierName . ' SPN'` for supplier fields (as seen in `getAvailableFieldTargets()` and `getSuggestedFieldMapping()`). This means LCSC supplier part numbers are mapped to the field name `LCSC SPN`, not `LCSC`.
The bug was that the code was checking for the wrong key name, which would cause LCSC supplier part numbers to be silently ignored and not included in the BOM entry comments.

This got fixed in my first commit.

In my second commit, I extended the unit tests to also check for `LCSC SPN` in the comment after import.

Commit 1 + 2 therefore fix a bug that currently exists in Part-DB.

I found it weird, that the logic for adding the SPN's to the comment only existed for LCSC.
In commit 3, I therefore modified my previous fix to dynamically work for all suppliers being configured.
In commit 4, I moved the logic of getting the supplies outside the loop to be only executed once.

If you want something changed, feel free to tell me!
If you prefer, I can move commit 3+4 into a separate PR, since they technically do not fix the bug but extend the features of LCSC to other suppliers.

### DIFF
--- a/src/Services/ImportExportSystem/BOMImporter.php
+++ b/src/Services/ImportExportSystem/BOMImporter.php
@@ -274,6 +274,13 @@ class BOMImporter
         $entries_by_key = []; // Track entries by name+part combination
         $mapped_entries = []; // Collect all mapped entries for validation
 
+        // Fetch suppliers once for efficiency
+        $suppliers = $this->entityManager->getRepository(\App\Entity\Parts\Supplier::class)->findAll();
+        $supplierSPNKeys = [];
+        foreach ($suppliers as $supplier) {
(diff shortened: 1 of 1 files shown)


jbtronics:
Looks good. Thanks.

PR #1209, opened 2026-01-25
add supplier SPN linking for BOM import

MayNiklas:
This pull request enhances the BOM import functionality, specifically improving how supplier part numbers (SPNs) are used to automatically link BOM entries to existing parts in the database.

### Improvements to BOM import logic:

* Added logic to attempt linking BOM entries to existing parts using supplier SPNs when a Part-DB ID is not provided. This checks all known suppliers and matches entries based on supplier name and SPN.
* Optimized supplier lookup by creating a mapping of supplier names to supplier objects for efficient SPN matching.

In my opinion, there is no reason not to match a part when a supplier SPN provided by the user does match to something already in the DB.
A lot of people have for example LCSC SPN's in their BOM's.
Automatically linking matched parts greatly simplifies the project creation workflow!
I for example would use this a lot even while still working on the project, to make sure I have all the needed parts early in the process!

If you have a different opinion on it, feel free to tell me :)
I'm willing to make changes if requested :)

Note: I'm not able to execute the PHP tests locally. While I did deploy the code on my server and it did work, I have not yet executed the tests. I therefore hope for them to pass on your CI :)

### DIFF
--- a/tests/Services/ImportExportSystem/BOMImporterTest.php
+++ b/tests/Services/ImportExportSystem/BOMImporterTest.php
@@ -616,6 +616,181 @@ class BOMImporterTest extends WebTestCase
         $this->assertEquals('R1,R2', $bom_entries[0]->getMountnames());
     }
 
+    public function testStringToBOMEntriesKiCADSchematicWithSupplierSPN(): void
+    {
+        // Create test supplier
+        $lcscSupplier = new Supplier();
+        $lcscSupplier->setName('LCSC');
+        $this->entityManager->persist($lcscSupplier);
+
+        // Create a test part with required fields
+        $part = new Part();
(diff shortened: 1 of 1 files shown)


dj-bauer:
This patch works on my instance and i tested it with both LCSC and Digikey.
I also tested the case where different two SPN linking to different Parts are assigned to the same BOM-Item and it seems to match one of the two, but the exact assignment still can be changed in the preview.
I like 👍🏼

jbtronics:
Looks good. Thank you.

Davis8483:
@jbtronics is there a way to get around the designator requirement? We use PartDB for our FTC robotics club, so we're hoping to simply export the BOM from Onshape and import into PartDB. Only SPN and quantity are needed.

PR #1211, opened 2026-01-26
Added Conrad provider

buergi:
I implemented the official Conrad API.
Unfortunately it seems it is currently not possible to officially get an API key, as to register at https://developer.conrad.com/ your mail needs to be allowlisted.
I tested the provider with the API keys from the conrad website.

The selected countries determines the shop instance, and the selected language needs to be supported by that instance. So for example CH supports DE or FR as languages but selecting country FR and language EN will fail to query most description and parameter information. Currently the provider does not check this.

The most important information is implemented: name, description, image, datasheet, price information, parameters.

### DIFF
--- a/docs/usage/information_provider_system.md
+++ b/docs/usage/information_provider_system.md
@@ -278,6 +278,19 @@ The following env configuration options are available:
 * `PROVIDER_BUERKLIN_CURRENCY`: The currency you want to get prices in if available (optional, 3 letter ISO-code, default: `EUR`).
 * `PROVIDER_BUERKLIN_LANGUAGE`: The language you want to get the descriptions in. Possible values: `de` = German, `en` = English. (optional, default: `en`)
 
+### Conrad
+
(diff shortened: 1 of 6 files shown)


jbtronics:
Funnily I started working on a conrad provider just yesterday: https://github.com/Part-DB/Part-DB-server/tree/conrad_provider

I will take a look how to merge both together.

buergi:
Ah so sorry, had it working already two or three weeks ago but procrastinated the PR for so long. But I'm glad you already looked into it, now you can probably better judge if my implementation is sane or if I overlooked something. At least in my tests it was working pretty well and I already imported multiple components into my personal database with it. Only added some better handling of countries and languages, hopefully I didn't introduce problems there.

Looking forward to seeing a Conrad provider in the next version.

jbtronics:
I used some logic from your PR to finish my provider version and merged it.
That makes this PR obsolete.

Still thank you very much.

Issue #603, opened 2024-04-21
Cost calculation after building project


MG-5:
First of all, your project is really great, practical and I'm glad that it makes my everyday assembly so much easier. 

It would be cool if, as soon as you build a project, the total price of all items for which a unit price exists could be calculated and displayed.

partywagon112:
Had the same thought when I imported my project. Looks like doing something like that is supported inthe API, but definitely would be amazing with a native indication price based on the item data.

mase76:
That would be a great feature. With that you could offer buildkits in your Hackerspace.

ehajo:
+1 :-)

chickey:
Just wanted to bump this up as i think it would be very useful and something i liked from parts box.  I don't mind it being worst case pricing without accommodating price breaks etc.  I'm a small maker so rarely have multiples of the same part in from different manufacture's so that would not be an issue either.  Partsbox also allowed you to input the quantity to get a rough bom cost.

![image](https://github.com/user-attachments/assets/407e7921-faff-44e1-93f9-44368be05ca5)

dtk1985:
+1

MelvinValerius:
+1

JochenVandorpe:
+1 (for now still tracking this in excel based on the schematic BOM but feels like overhead when creating the projects in part-DB to 'book' them there for parts count update)

Stephane-Peel:
+1

PR #1221, opened 2026-02-02
Add Price and Extended price to Project Bom

whinis:
This seems to be one of the most requested changes and I also needed this particular feature. 
requested in #603  and others in discussion. So this allows one to see the price of each item in the bom if it exists in any way. It currently selects just the first supplier and the first (likely most expensive) unit price for the part.

### DIFF
--- a/src/DataTables/ProjectBomEntriesDataTable.php
+++ b/src/DataTables/ProjectBomEntriesDataTable.php
@@ -39,6 +39,7 @@ use Omines\DataTablesBundle\Column\TextColumn;
 use Omines\DataTablesBundle\DataTable;
 use Omines\DataTablesBundle\DataTableTypeInterface;
 use Symfony\Contracts\Translation\TranslatorInterface;
+use Brick\Math\BigDecimal;
 
 class ProjectBomEntriesDataTable implements DataTableTypeInterface
 {
@@ -178,6 +179,67 @@ class ProjectBomEntriesDataTable implements DataTableTypeInterface
 
                     return '';
                 }
+            ])
+            ->add('price', TextColumn::class, [
+                'label' => 'project.bom.price',
+                'render' => function ($value, ProjectBOMEntry $context) {
+                    // Let's attempt to get the part, if we don't we will just assume zero
+                    $part = $context->getPart();
+                    $price = BigDecimal::zero();
+                    $pricedetails = null;
+                    $order = null;
+                    // Check if we get a part and get the first order if so
+                    // if not see if there is a non-part price set
+                    if($part) {
+                        $order = $context->getPart()->getOrderdetails()->first();
+                    } else if($context->getPrice() !== null) {
+                        $price = $context->getPrice();
+                    }
+
+                    // check if there is an order, if so get the first pricedetail of the order
+                    if($order!==null && $order !== false) {
+                        $pricedetails = $order->getPricedetails()->first();
+                    }
+
+                    // check if there is a pricedetail, if so get the first price
+                    if($pricedetails !== null && $pricedetails !== false) {
+                        $price = $pricedetails->getPrice();
+                    }
+
+                    // return the price
+                    return  htmlspecialchars(number_format($price->toFloat(),2));
+                },
+                'visible' => false,
+            ])
+            ->add('ext_price', TextColumn::class, [
+                'label' => 'project.bom.ext_price',
+                'render' => function ($value, ProjectBOMEntry $context) {
+                    // Let's attempt to get the part, if we don't we will just assume zero
+                    $part = $context->getPart();
+                    $price = BigDecimal::zero();
+                    $pricedetails = null;
+                    $order = null;
+                    // Check if we get a part and get the first order if so
+                    // if not see if there is a non-part price set
+                    if($part) {
+                        $order = $context->getPart()->getOrderdetails()->first();
+                    } else if($context->getPrice() !== null) {
+                        $price = $context->getPrice();
+                    }
+
+                    // check if there is an order, if so get the first pricedetail of the order
+                    if($order!==null && $order !== false) {
+                        $pricedetails = $order->getPricedetails()->first();
+                    }
+
+                    // check if there is a pricedetail, if so get the first price
+                    if($pricedetails !== null && $pricedetails !== false) {
+                        $price = $pricedetails->getPrice();
+                    }
+
+                    // return the price
+                    return  htmlspecialchars(number_format($price->toFloat() * $context->getQuantity(),2));
+                },
             ])
 
             ->add('addedDate', LocaleDateTimeColumn::class, [
--- a/src/DataTables/ProjectBomEntriesDataTable.php
+++ b/src/DataTables/ProjectBomEntriesDataTable.php
@@ -32,6 +32,7 @@ use App\Entity\ProjectSystem\ProjectBOMEntry;
 use App\Services\ElementTypeNameGenerator;
 use App\Services\EntityURLGenerator;
 use App\Services\Formatters\AmountFormatter;
+use App\Services\Parts\PricedetailHelper;
 use Doctrine\ORM\QueryBuilder;
 use Omines\DataTablesBundle\Adapter\Doctrine\ORM\SearchCriteriaProvider;
 use Omines\DataTablesBundle\Adapter\Doctrine\ORMAdapter;
@@ -44,7 +45,8 @@ use Brick\Math\BigDecimal;
 class ProjectBomEntriesDataTable implements DataTableTypeInterface
 {
     public function __construct(protected TranslatorInterface $translator, protected PartDataTableHelper $partDataTableHelper,
-        protected EntityURLGenerator $entityURLGenerator, protected AmountFormatter $amountFormatter)
+        protected EntityURLGenerator $entityURLGenerator, protected AmountFormatter $amountFormatter,
+        protected PricedetailHelper $pricedetailHelper)
     {
     }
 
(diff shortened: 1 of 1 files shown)


jbtronics:
Thanks,

The PricedetailHelper has an calculateAvg() price to calculate an average price across all suppliers, it also handles currency conversions correctly.
Also you should use the correct Intl Number Formatter to get the correctly formatted price for users locale. The currency symbol should be included, to mark which currency this is.

Unfortunatley the whole price infrastructure in Part-DB is quite limited and ideally would need some heavy rewrites, with some more powerful functions to make things like this easier...

MayNiklas:
> The PricedetailHelper has an calculateAvg() price to calculate an average price across all suppliers, it also handles currency conversions correctly.

Just a random thought:
When calculating the price of a project, I would prefer the price being worse compared to it being too good.
This way, we don't run into the risk of thinking something is cheaper than it actually is.

The complicated thing about pricing: it won't reflect the actual price we bought parts for & sometimes our stock even might contain the same part bought for different prices.

I would love the price estimation to reflect what I'm able to buy the parts now for:
Instead of using the average price, which might result in a price being too good, I would prefer to use the prices of the amounts we actually require to build a fixed amount of units.

Long story short:
I don't know how others think about it, but I would prefer prices to be higher compared tp them being too low.
I would even be fine with PartDB just using the more expensive prices for lower amounts of parts, instead of using the price average.

I know prices in PartDB are complicated!
Of any specific help is required, I would love to help!

whinis:
> > The PricedetailHelper has an calculateAvg() price to calculate an average price across all suppliers, it also handles currency conversions correctly.
> 
> Just a random thought: When calculating the price of a project, I would prefer the price being worse compared to it being too good. This way, we don't run into the risk of thinking something is cheaper than it actually is.
> 
> The complicated thing about pricing: it won't reflect the actual price we bought parts for & sometimes our stock even might contain the same part bought for different prices.
> 
> I would love the price estimation to reflect what I'm able to buy the parts now for: Instead of using the average price, which might result in a price being too good, I would prefer to use the prices of the amounts we actually require to build a fixed amount of units.
> 
> Long story short: I don't know how others think about it, but I would prefer prices to be higher compared tp them being too low. I would even be fine with PartDB just using the more expensive prices for lower amounts of parts, instead of using the price average.
> 
> I would love the price estimation to reflect what I'm able to buy the parts now for

Thats what the PR essentially does or theoretically does. I could force it but it currently gets the first price which atleast in my setup tends to be qty 1 price.

MayNiklas:
@whinis I implemented the changes @jbtronics asked for.
Should I share them with you so you can commit them, adding me as a co-author?

MayNiklas:
My previous comment can be ignored - I realised how average price works! :)

MayNiklas:
@whinis If you want to, you can commit my patch integrating the requested changes.
Save the following diff as a file `commit.diff`.
Then use `git am --3way commit.diff`.
`git am` is used to mailbox patches :)

The changes greatly simplify the price calculation!
I have to agree with @jbtronics - this is much nicer!

```diff
From 44a9aa16b312f64b83777efd8c369a8ce3805310 Mon Sep 17 00:00:00 2001
From: MayNiklas <info@niklas-steffen.de>
Date: Fri, 13 Mar 2026 21:15:25 +0100
Subject: [PATCH] project prices: use calculateAvg


diff --git a/src/DataTables/ProjectBomEntriesDataTable.php b/src/DataTables/ProjectBomEntriesDataTable.php
index 3acef03f..ec1f5ff4 100644
--- a/src/DataTables/ProjectBomEntriesDataTable.php
+++ b/src/DataTables/ProjectBomEntriesDataTable.php
@@ -39,12 +39,14 @@ use Omines\DataTablesBundle\Column\TextColumn;
 use Omines\DataTablesBundle\DataTable;
 use Omines\DataTablesBundle\DataTableTypeInterface;
 use Symfony\Contracts\Translation\TranslatorInterface;
+use App\Services\Parts\PricedetailHelper;
 use Brick\Math\BigDecimal;
 
 class ProjectBomEntriesDataTable implements DataTableTypeInterface
 {
     public function __construct(protected TranslatorInterface $translator, protected PartDataTableHelper $partDataTableHelper,
-        protected EntityURLGenerator $entityURLGenerator, protected AmountFormatter $amountFormatter)
+        protected EntityURLGenerator $entityURLGenerator, protected AmountFormatter $amountFormatter,
+        protected PricedetailHelper $pricedetailHelper)
     {
     }
 
@@ -183,62 +185,18 @@ class ProjectBomEntriesDataTable implements DataTableTypeInterface
             ->add('price', TextColumn::class, [
                 'label' => 'project.bom.price',
                 'render' => function ($value, ProjectBOMEntry $context) {
-                    // Let's attempt to get the part, if we don't we will just assume zero
-                    $part = $context->getPart();
-                    $price = BigDecimal::zero();
-                    $pricedetails = null;
-                    $order = null;
-                    // Check if we get a part and get the first order if so
-                    // if not see if there is a non-part price set
-                    if($part) {
-                        $order = $context->getPart()->getOrderdetails()->first();
-                    } else if($context->getPrice() !== null) {
-                        $price = $context->getPrice();
-                    }
-
-                    // check if there is an order, if so get the first pricedetail of the order
-                    if($order!==null && $order !== false) {
-                        $pricedetails = $order->getPricedetails()->first();
-                    }
-
-                    // check if there is a pricedetail, if so get the first price
-                    if($pricedetails !== null && $pricedetails !== false) {
-                        $price = $pricedetails->getPrice();
-                    }
+                    $price = $this->getBomEntryUnitPrice($context);
 
-                    // return the price
-                    return  htmlspecialchars(number_format($price->toFloat(),2));
+                    return htmlspecialchars(number_format($price->toFloat(), 2));
                 },
                 'visible' => false,
             ])
             ->add('ext_price', TextColumn::class, [
                 'label' => 'project.bom.ext_price',
                 'render' => function ($value, ProjectBOMEntry $context) {
-                    // Let's attempt to get the part, if we don't we will just assume zero
-                    $part = $context->getPart();
-                    $price = BigDecimal::zero();
-                    $pricedetails = null;
-                    $order = null;
-                    // Check if we get a part and get the first order if so
-                    // if not see if there is a non-part price set
-                    if($part) {
-                        $order = $context->getPart()->getOrderdetails()->first();
-                    } else if($context->getPrice() !== null) {
-                        $price = $context->getPrice();
-                    }
-
-                    // check if there is an order, if so get the first pricedetail of the order
-                    if($order!==null && $order !== false) {
-                        $pricedetails = $order->getPricedetails()->first();
-                    }
+                    $price = $this->getBomEntryUnitPrice($context);
 
-                    // check if there is a pricedetail, if so get the first price
-                    if($pricedetails !== null && $pricedetails !== false) {
-                        $price = $pricedetails->getPrice();
-                    }
-
-                    // return the price
-                    return  htmlspecialchars(number_format($price->toFloat() * $context->getQuantity(),2));
+                    return htmlspecialchars(number_format($price->toFloat() * $context->getQuantity(), 2));
                 },
             ])
 
@@ -268,6 +226,16 @@ class ProjectBomEntriesDataTable implements DataTableTypeInterface
         ]);
     }
 
+    private function getBomEntryUnitPrice(ProjectBOMEntry $entry): BigDecimal
+    {
+        if ($entry->getPart() instanceof Part) {
+            return $this->pricedetailHelper->calculateAvgPrice($entry->getPart(), $entry->getQuantity()) ?? BigDecimal::zero();
+        }
+
+        return $entry->getPrice() ?? BigDecimal::zero();
+    }
+
+
     private function getQuery(QueryBuilder $builder, array $options): void
     {
         $builder->select('bom_entry')
```

whinis:
@MayNiklas Alright I incorporated that. It does look nicer in this case.

MayNiklas:
@jbtronics this PR should be ready for a re-review :)

MayNiklas:
@jbtronics I just noticed, that `pricedetailHelper->calculateAvgPrice` does not work as expected for smaller quantities, when quantity < minimum order amount.
In those cases, the function returns NULL, leading to a price of 0,00 being used.
Would it be possible to fix that behaviour? Or does anything speak against it?

src/DataTables/ProjectBomEntriesDataTable.php
```diff
@@ -205,6 +227,14 @@ function (QueryBuilder $builder) use ($options): void {
             ],
         ]);
     }
+    private function getBomEntryUnitPrice(ProjectBOMEntry $entry): BigDecimal
+    {
+        if ($entry->getPart() instanceof Part) {
+            return $this->pricedetailHelper->calculateAvgPrice($entry->getPart(), $entry->getQuantity()) ?? BigDecimal::zero();
+        }
+
+        return $entry->getPrice() ?? BigDecimal::zero();
```
MayNiklas: We previously not correctly handled cases with amount < minimum order amount.
This should fix it!

```suggestion
        if ($entry->getPart() instanceof Part) {
            $amount = $entry->getQuantity();
            $minOrderAmount = $this->pricedetailHelper->getMinOrderAmount($entry->getPart());
            if ($minOrderAmount !== null) {
                $amount = max($amount, $minOrderAmount);
            }
            return $this->pricedetailHelper->calculateAvgPrice($entry->getPart(), $amount) ?? BigDecimal::zero();
        }
        return $entry->getPrice() ?? BigDecimal::zero();
```

MayNiklas (commented):
I just added two suggestions I noticed this morning :)
- prices should maybe be rounded up, since 0,00€ parts are realistically more 0,01€ stan 0,00€
- in cases where getMinOrderAmount > amount, we previously returned NULL. With my suggestion we would continue to work with the minimum ammount that can be ordered.

src/DataTables/ProjectBomEntriesDataTable.php
```diff
@@ -205,6 +227,14 @@ function (QueryBuilder $builder) use ($options): void {
...
+    }
```
MayNiklas: I just forgot to add roundUpDisplayedPrice
```suggestion
    private function roundUpDisplayedPrice(BigDecimal $price): BigDecimal
    {
        return $price->toScale(2, RoundingMode::UP);
    }
```

jbtronics:
This PR became obsolete by PR #1345

PR #1224, opened 2026-02-04
Add simple buttons to use builtin export functions

whinis:
Another heavily requested feature with a quick fix. This just added the standard export for csv and excel built into datatables bundles. Need some styling and a way to add translations.


jbtronics:
I am not sure what the advantages of this should be over the "normal" export function, which is done on server side?
Especially as the client side export is quite limited in the data it has access too, and it is not necesaarily stable in its output.

whinis:
> I am not sure what the advantages of this should be over the "normal" export function, which is done on server side? Especially as the client side export is quite limited in the data it has access too, and it is not necesaarily stable in its output.

Is there a server side export for a BOM? I couldn't find it in the code and this was sufficient for my needs to create an order list. Obviously a more detailed server side exporter to allow selection for a particular provider would be nice but was out of the scope of this work.

Issue #536, opened 2024-03-03
Select Fields which appear in KiCad


foreverimagining:
**Is your feature request related to a problem? Please describe.**
I found this project just a couple days ago because I was hoping for a way to store information about components that I select and/or have in stock that would be easily imported into KiCad so I don't have to manually input data for each part every time I make a new project or copy-and-paste from one project to another. It does a great job at that, and KiCad 8 is fantastic as well. However, since I plan to share some of my projects as open-source hardware, I want to make it easy for people to identify the specific components I use at common distributors by including their particular part numbers. Unfortunately, only a few fields are actually included in the information passed to KiCad, and I don't consider the manufacturer's part number to be quite sufficient when it can sometimes turn up alternatives in searches.

**Describe the solution you'd like**
I would like to be able to select which fields are exposed to KiCad besides the standards that are already included. This should include selecting parameters as well as distributor information. I mostly want to be able to add distributor part numbers which should have field names associating them with the distributor, but I can foresee cases where I might want to include important parameters that might have an impact on component selection in the projects. For implementation in the UI, perhaps a checkmark field next to each parameter and distributor information would be the simplest way to toggle inclusion? If this is included on the pricing information, we could even include the price for the distributor we would be buying from (likely the lowest cost option that fulfills the number of pieces needed).

**Describe alternatives you've considered**
Unless/until this feature is added, I figure I will just have to input the distributor part numbers manually. It won't be completely awful if I wait until the end of the project when all of the components are decided and then use the bulk-edit feature in conjunction with Part-DB as a reference. It's still a time-saver as-is.

Thank you so much for all the work you've done on this. It really solves a problem for me, and I look forward to seeing if I can integrate this into my Grocy and/or Home Assistant instances using the API for automation when I have more time in the future.

Kater--S:
Same here. I would like to use the KiCad plugin [(JLCPCB) Fabrication Toolkit](https://github.com/bennymeg/Fabrication-Toolkit) by Benny Megidish in combination with PartDB. The plugin needs the LCSC part number in order to build the production artefacts for JLCPCB. The needed value is readily available in PartDB under Purchase information → Supplier LCSC → Supplier part number, but it is not visible from KiCad. Ideally the target key name for such a field would be configurable, but if this is too complicated it would be sufficient to have it available in any identifiable way at all, maybe requiring a small change to the plugin's source.

PR #1232, opened 2026-02-08
Enhance KiCad API with stock quantity, datasheet URL, and HTTP caching

Sebbeben:
## Summary

Adds three improvements to the KiCad HTTP Library API (\):

- **Stock quantity field** — Exposes total in-stock quantity as a \ field and storage locations as \. Excludes expired lots and lots with unknown quantity. Engineers can now see parts availability directly in KiCad's symbol chooser.
- **Datasheet URL resolution** — The \ field now resolves to an actual datasheet file URL instead of always using the Part-DB page link. Searches attachments by type name containing "datasheet", then by attachment name, then falls back to the first PDF attachment, and finally to the Part-DB page URL.
- **HTTP cache headers** — Adds \ and \ headers to all API responses. Supports conditional requests (\) returning \ when data hasn't changed. Categories/part lists cache for 5 minutes, part details for 1 minute.

Also fixes a pre-existing bug in \ which was actually testing Part 1 instead of Part 2.

## Related issues

- Addresses discussion https://github.com/Part-DB/Part-DB-server/discussions/395 (datasheet URL passthrough)
- Partially addresses #536 (configurable field exposure — supplier part numbers are now exposed)
- Stock quantity was a top community request for the KiCad integration

## New fields exposed

| Field | Visibility | Description |
|-------|-----------|-------------|
| \ | Hidden | Total quantity across all valid (non-expired, known) lots |
| \ | Hidden | Comma-separated list of storage locations with stock |
| \ | Hidden | Direct link to the part's info page in Part-DB |

## Test plan

- [x] Added 14 unit tests in \ covering stock calculation (zero, single, multiple lots, expired/unknown exclusion), datasheet resolution (type name, attachment name, PDF fallback, space variant), storage locations, and field visibility
- [x] Added HTTP cache header tests and 304 conditional request test in - [x] Fixed \ to test actual Part 2 with EDA inheritance from category/footprint
- [x] Manually tested all three features on a live Part-DB instance with KiCad 9

### DIFF
--- a/src/Controller/KiCadApiController.php
+++ b/src/Controller/KiCadApiController.php
@@ -92,17 +92,12 @@ class KiCadApiController extends AbstractController
      * Creates a JSON response with HTTP cache headers (ETag and Cache-Control).
      * Returns 304 Not Modified if the client's ETag matches.
      */
-    private function createCachedJsonResponse(Request $request, array $data, int $maxAge): JsonResponse
+    private function createCachedJsonResponse(Request $request, array $data, int $maxAge): Response
     {
-        $etag = '"' . md5(json_encode($data)) . '"';
-
-        if ($request->headers->get('If-None-Match') === $etag) {
-            return new Response('', Response::HTTP_NOT_MODIFIED);
-        }
-
         $response = new JsonResponse($data);
+        $response->setEtag(md5(json_encode($data)));
         $response->headers->set('Cache-Control', 'private, max-age=' . $maxAge);
-        $response->headers->set('ETag', $etag);
+        $response->isNotModified($request);
 
         return $response;
     }


jbtronics:
From some quick look:

* I thing the behavior what gets exposed as datasheet URL should be configurable. A system setting that switches between yours and the old behavior would be good, as I think some people value being able to jump to Part-DB directly (and at least in my experiments the datasheet field was the only one which gets rendered as link)

Sebbeben:
Thanks for the feedback! I've added a configurable system setting "Datasheet field links to PDF" (`Settings → KiCad/EDA`) that switches between the new PDF URL behavior and the old Part-DB page link. It defaults to enabled (PDF) but users can toggle it. Also configurable via `EDA_KICAD_DATASHEET_AS_PDF` env var.

I've opened a more comprehensive PR #1241 that includes this fix along with several other KiCad improvements (API v2 with volatile fields, batch EDA editing, configurable field export, BOM multi-footprint merging, etc.). Closing this PR in favor of #1241.

PR #1240, opened 2026-02-11
Autofocus fields

### LINKED ISSUE #1157: Auto-focus main fields on page load
It would be great to auto-focus the main fields on some pages, for example:
- new part would focus the name input
- on add/withdrawn stock qty would focus Amount
- when adding a new stock the amount field would get focused probably too
etc.

it would make things easiers and faster, just click + type a number and enter without having to switch to the mouse just to focus the field.

d-buchmann:
Fixes #1157.
- Focus `name` field on all new entity editors
- Focus `amount` on add/withdraw modal
- Focus first `number` type input on any newly added CollectionType table row
(debatable and slightly buggy - focuses `Max` field in parameters...)
- Nothing for Project BOM yet; could be useful there too

It would be even more favorable if the user could configure if they want to use autofocus and/or for which fields/dialogs it should be enabled. I read that autofocus can cause issues with accessibility, especially when using screenreaders. Unfortunately I have no experience in that area :/

### DIFF
--- a/assets/controllers/elements/collection_type_controller.js
+++ b/assets/controllers/elements/collection_type_controller.js
@@ -79,9 +79,13 @@ export default class extends Controller {
         //Afterwards return the newly created row
         if(targetTable.tBodies[0]) {
             targetTable.tBodies[0].insertAdjacentHTML('beforeend', newElementStr);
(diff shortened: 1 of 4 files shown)


jbtronics:
Thanks

PR #1241, opened 2026-02-11
Enhance KiCad integration: API v2, batch EDA editing, field export control

### LINKED ISSUE #536: Select Fields which appear in KiCad

### LINKED ISSUE #554: Mass editing of parts
Hello,
a feature for mass editing parts would be really nice.
It's especially usefull for new functions like EDA information.
For example editing the eda information of all of my 0603 MLCCs.

When selecting multiple parts in an list, a button "Edit Selected" where values of all parts can be overwritten.

**Additional context**

### LINKED ISSUE #614: Better import feature
I'm trying to move my existing KiCad database to Part-DB via Import Data feature. But I found it can't import all parameters that a part contain. Basic like its value, case and EDA information.

Allow direct field createion when import or more allowed data field like EDA information.

Or do it in a multi step process.
1. Import parts as current
2. Create parameters in parts in WebGUI.
3. Export a CSV of All parts with parameters field. Preferable in OR logic.
So part X have parameter A, B. Part Y have parameter B, C. 
The output CSV have A,B,C field.
4. Edit externally.
5. Import the CSV back and update all the field

**Additional context**
I believe the current version support 1,2 and 3 already. So it is up to how to enhance the CSV import function

### LINKED ISSUE #778: KiCAD BOM importer fails with project where part has multiple footprints
**Describe the bug**
KiCAD's BOM import feature seems to break down whenever a board is using components for which there are multiple footprints used on the board as I can have only unique components in the PartDB BOM.

For example, if I have a board which has multiple 4.7uF capacitors, but I use at least two different footprints for placement:
```csv
4;"C7,C6,C12";"C_0805_2012Metric_Pad1.15x1.40mm_HandSolder";3;"4.7uF";;;
19;"C16,C15,C10,C21";"C_0805_2012Metric_Pad1.18x1.45mm_HandSolder";4;"4.7uF";;;
```

**To Reproduce**
1. Fetch the BlueSCSI 50pin desktop project from https://github.com/BlueSCSI/BlueSCSI-v2/tree/main/hardware/Desktop_50_Pin and generate BOM in the PCB editor
2. Import BOM...
3. Assign that 4.7uF component to lines 4 and 19
4. Attempt to save...

**Expected behavior**
Not really sure what would be good solution here...

**Screenshots**

**Server Side**
- Part-DB Version: [v1.14.4](https://github.com/Part-DB/Part-DB-server/releases/tag/v1.14.4)
- PHP Version: (the one that comes with docker image)
- Database Server: PostgreSQL

Sebbeben:
## Summary

Comprehensive improvements to the KiCad HTTP Library API and EDA workflow, addressing several open issues.

### KiCad API Enhancements
- **Stock quantity & storage location** fields in API response
- **Datasheet PDF URL**: Resolves actual PDF attachments instead of linking to Part-DB page (#395)
- **HTTP cache headers**: ETag + Cache-Control for conditional requests
- **`?minimal=true` query param**: Faster category part loading for large categories
- **Category descriptions**: Uses category comment when available
- **KiCad API v2** (`/kicad-api/v2/`): Adds volatile field support — Stock and Storage Location are shown in KiCad but NOT saved to schematic files

### Configurable Field Export (#536)
- **Parameter `kicad_export` flag**: Per-parameter checkbox to control which specifications are exposed to KiCad (e.g., tolerance, voltage rating)
- **Orderdetail `kicad_export` flag**: Per-supplier control of which SPNs get exported. Backward compatible — if no checkbox is checked, all SPNs export as before.

### CSV Import of EDA Fields (#614)
- Support for `kicad_symbol`, `kicad_footprint`, `kicad_reference`, `kicad_value`, `eda_visibility`, `eda_exclude_bom`, `eda_exclude_board`, `eda_exclude_sim` columns in CSV part import
- Enables bulk migration from existing KiCad database libraries

### Batch EDA Editing (#554)
- New batch action in parts table to edit EDA/KiCad fields for multiple selected parts
- Supports: reference prefix, value, symbol, footprint, visibility, and exclude flags
- Pre-populates shared values when all selected parts have the same value

### BOM Import: Multi-Footprint Merging (#778)
- When a BOM entry links to a Part-DB part (via ID or SPN), entries are merged by part ID regardless of footprint variant
- Example: Same capacitor with standard and hand-solder pads merges into one BOM entry

### EDA Status Indicator
- New "EDA" column in parts datatable showing completeness at a glance
- Green bolt = symbol + footprint + reference all set; Yellow bolt = partial
- Clickable — opens directly to the part's EDA settings tab

### Console Command: `partdb:kicad:populate`
- Bulk-assigns KiCad symbol/footprint paths based on category/footprint mappings
- Supports CSV mapping files, dry-run mode, and regex patterns
- Useful for initial KiCad integration setup on large databases

### Code Quality
- Deduplicated KiCost supplier field generation (single loop instead of two)
- Fixed redundant `getEdaInfo()` call in visibility check
- Moved `manf`/`manf#` fields outside orderdetails conditional (always present)
- Full PHPStan compliance, 1216 PHPUnit tests passing

## Related Issues
- Closes #536 (Configurable field exposure)
- Closes #614 (CSV import EDA fields)
- Closes #554 (Batch EDA editing)
- Closes #778 (BOM multi-footprint merging)
- Addresses #395 (Datasheet PDF URL)

## Database Migrations
- `Version20260208190000`: Adds `kicad_export` column to `parameters` table
- `Version20260210120000`: Adds `kicad_export` column to `orderdetails` table

## Test Plan
- [x] PHPStan: 0 errors (685 files)
- [x] PHPUnit: 1216 tests, 3080 assertions — all pass
- [x] New tests: KiCadHelperTest (unit), KiCadApiControllerTest (functional), PopulateKicadCommandTest, PartNormalizerTest
- [x] Manual UI testing: EDA status column, batch editing, orderdetail checkbox, BOM import with multi-footprint CSV
- [x] API testing: v1 and v2 endpoints verified with Bearer token auth


jbtronics:
Looks interesting thanks.

Some remarks:
* The database migrationss should be merged

## Kicad populate:

* The populate kicad command, should be documented. List them at the console commands list docs and explain its use in the kicad docs page.
* It is probably useful to check for the alternative names of categories and footprints too, if there are matches with the known list
* I not sure if the current model with the mapping list hardcoded in the command is flexible enough for reality. Maybe it would make sense, to allow loading CSV or JSON file with the mapping, so that can be overwritten/extended by users. And supplying a default mapping list with Part-DB that is used, if there is no user supplied one. But this is probably not that important for start

## Kicad API Controller
* `createCacheableJsonResponse` is maybe a better fitting name for the function instead of `createCachedJsonResponse`
* Have you tried if these caching tags makes a big difference in actual performance? I would have expected the limiting factor is the response time of symfony, not the data transfer.

## KiCad v2 api
* Do you have a link to the KiCAD  v2 API documentation?  

## Abstract parameter & orderdetail
* Maybe `eda_visibility` is a better name than `kicad_export`.
* It might make sense to have an system option to set the default value for these. Either for newly created elements, or for all elements, where the value is not explictly set

## Parameter type
* The EDA export field should only be visible for parameters on parts, for other entities it has no function I think.

## KiCAD Helper
* For future proofing it is probably useful to not use a bool for switching between the different API version, but an integer number (with a check for our supported versions). Or use separate methods like `getPartsV1()` and `getPartsV2()`
* It seems the `manf` field for KiCost was removed from the API response.

Sebbeben:
Thanks a lot for the thorough review and the great suggestions! Really appreciate you taking the time — several of these ideas (especially `eda_visibility` naming and the system-wide default) are much better than what I had.

Here's my plan for addressing everything:

**Will implement:**
- Merge the two database migrations into one
- Rename `kicad_export` → `eda_visibility` everywhere (entities, forms, templates, translations, tests, migration column name)
- Rename `createCachedJsonResponse` → `createCacheableJsonResponse`
- Change `bool $apiV2` to `int $apiVersion` with supported version validation
- Make the EDA export field only visible for parameters on parts (not other entity types)
- PopulateKicadCommand: also check alternative names of categories/footprints for matches against the mapping list
- PopulateKicadCommand: support loading an external JSON mapping file (ship a default one with Part-DB)
- Add system-wide default setting for `eda_visibility` (make the field nullable, `null` = use system default)
- Document the populate command in the console commands list and KiCad docs page

**Regarding the `manf` field:** I believe it's still there — I moved it earlier in the method (before the orderdetails loop, under "always present, independent of orderdetails") so it's no longer tied to the supplier filtering logic. But the diff might make it look like a removal if you only see the old location. Please let me know if I'm missing something!

**Regarding KiCad v2 API documentation:**
The v2 spec with volatile fields is documented here, though it's still in a development branch and not yet part of the official KiCad docs:
- https://gitlab.com/RosyDev/kicad-dev-docs/-/blob/http-lib-v2/content/apis-and-binding/http-libraries/http-lib-v2-00.adoc

The official v1 docs are at:
- https://dev-docs.kicad.org/en/apis-and-binding/http-libraries/index.html

**Question about v2:** Since the KiCad HTTP Library v2 spec is still in a development branch and not finalized yet, do you think we should hold off on the v2 API controller until the spec is stable? Or is it fine to ship as-is? Happy to defer the v2 parts to a follow-up PR if you prefer — just let me know.

jbtronics:
I don't really know how mature the V2 specs are so far... I mean it's not that problematic to include a preliminary support of that into Part-DB, but someone would need to finalize it, when the final specs are released.

Also as far as I understand the API docs, the root endpoint should return links to the other endpoints.

So I think it might be useful to split this into a separate PR, to discuss that further...

Sebbeben:
That makes a lot of sense, thank you for the suggestion! I've split the v2 API into a separate PR: #1260

This PR (#1241) now only contains the v1 improvements (eda_visibility rename, batch EDA editing, populate command enhancements, default EDA visibility setting, etc.) without any v2-specific code.

The v2 PR (#1260) depends on this one being merged first and includes:
- The v2 controller with volatile field support (Stock/Storage Location)
- Root endpoint now returns links to the categories endpoint (as you noted the API spec requires)
- `int $apiVersion` parameter on `KiCadHelper::getKiCADPart()` for clean version switching

I agree the v2 spec needs to be finalized before we commit fully to it. Happy to adjust the implementation once the spec is settled.

jbtronics:
thanks.

PR #1254, opened 2026-02-17
add option to disable keybindings fixing #1251

### LINKED ISSUE #1251: Add config option to disable keybindings for special characters
My current issue:
I'm on a MacBook with a German keyboard layout.
alt + 8 is used to insert {.
alt + 9 is used to insert }.
I noticed, that instead of writing `{}` I'm always writing `©{¥}` within the PartDB search & scanner.
I found https://docs.part-db.de/usage/keybindings.html and realised, that this is indeed a intended feature.
The annoying part: {} are part of many QR codes handed out by suppliers. I can't scan those barcodes correctly, since partdb is replacing my brackets with the wrong characters.

I'm hoping to find a solution, that key combinations used to present common characters on different operating systems are not used for keybindings.
A alternative would be for the keybindings to be configurable.

The easy approach I think would be easiest to quickly implement: adding a user option, to disable keybindings. I never intentionally used them, so I wouldn't mind to have them disabled for my user.
This would solve my issue without changing expected behaviour for users that depend on them.

MayNiklas:
This would fix #1251 
The keybindings conflict with important characters on Mac keyboards.
Currently for example `{}` always get replaced with ´©{¥}`, especially problematic in workflows using barcode scanners.


jbtronics:
Thanks.

Ideally this should be configurable on a per-user level, but this way is fine for now.
I plan to extend the settings-bundle for user based settings in the future, then this should be an easy extension.

PR #1255, opened 2026-02-17
Add manual backup creation and delete buttons to Update Manager

Sebbeben:
## Summary

Adds backup management buttons to the Update Manager and improves Docker support, with security hardening for sensitive operations.

### Backup Management
- **"Create Backup" button** in the backups tab for on-demand backups
- **Download button** for each backup — essential for Docker users since `var/backups/` is not a persistent volume
- **Trash icon delete buttons** for update logs and backup files
- **Docker support**: Backup creation and restore are now available for all installation types, not just git-based installs
- **Docker warning**: Info banner when Docker is detected, advising to download backups or mount `var/backups/`
- **SQLite restore fix**: Restore now uses the actual configured `DATABASE_URL` path instead of hardcoded `var/app.db`

### Security Hardening
- **`DISABLE_BACKUP_DOWNLOAD` env var** (default: `1` = disabled) — backups contain sensitive data (password hashes, secrets), so download is opt-in
- **Password confirmation** required before downloading a backup
- **CSRF token validation** on all mutating endpoints (create, delete, download, restore, start update)
- **`IS_AUTHENTICATED_FULLY`** check on all critical endpoints to prevent remember-me token abuse
- **Per-backup modals** for CSP/Turbo compatibility (no inline JavaScript)

### New `.env` variable

```env
# Disable backup download from the Update Manager UI (0=enabled, 1=disabled).
# Backups contain sensitive data including password hashes and secrets.
# When enabled, users must confirm their password before downloading.
DISABLE_BACKUP_DOWNLOAD=1
```

## Changes

| File | What changed |
|------|-------------|
| `UpdateManagerController.php` | Added `UserPasswordHasherInterface`, `DISABLE_BACKUP_DOWNLOAD` env var, `IS_AUTHENTICATED_FULLY` on critical endpoints, POST download with CSRF + password |
| `BackupManager.php` | Fixed SQLite restore to use Doctrine's configured path |
| `index.html.twig` | Per-backup download modals with password confirmation, Docker warning, `data-turbo="false"` for downloads |
| `.env` | Added `DISABLE_BACKUP_DOWNLOAD=1` default |
| `messages.en.xlf` | Added translation strings for download, password confirmation, security warning |

## Testing done

- Verified `DISABLE_BACKUP_DOWNLOAD=1` (default) hides download buttons entirely
- Verified `DISABLE_BACKUP_DOWNLOAD=0` shows download button with password modal
- Wrong password → flash error, download denied
- Correct password → backup ZIP downloaded, modal closes
- Create/delete backup and log with CSRF protection
- No CSP console errors (no inline JavaScript)
- All CI checks pass (PHPUnit 12/12, PHPStan, codecov)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -144,6 +144,18 @@ bundled with Part-DB. Set `DATABASE_MYSQL_SSL_VERIFY_CERT` if you want to accept
 * `ALLOW_EMAIL_PW_RESET`: Set this value to true, if you want to allow users to reset their password via an email
   notification. You have to configure the mail provider first before via the MAILER_DSN setting.
 
+### Update manager settings
+* `DISABLE_WEB_UPDATES` (default `1`): Set this to 0 to enable web-based updates. When enabled, you can perform updates
+  via the web interface in the update manager. This is disabled by default for security reasons, as it can be a risk if
+  not used carefully. You can still use the CLI commands to perform updates, even when web updates are disabled.
+* `DISABLE_BACKUP_RESTORE` (default `1`): Set this to 0 to enable backup restore via the web interface. When enabled, you can
+  restore backups via the web interface in the update manager. This is disabled by default for security reasons, as it can
+  be a risk if not used carefully. You can still use the CLI commands to perform backup restores, even when web-based
+  backup restore is disabled.
+* `DISABLE_BACKUP_DOWNLOAD` (default `1`): Set this to 0 to enable backup download via the web interface. When enabled, you can download backups via the web interface
+  in the update manager. This is disabled by default for security reasons, as it can be a risk if not used carefully, as
+  the downloads contain sensitive data like password hashes or secrets.
+
 ### Table related settings
 
 * `TABLE_DEFAULT_PAGE_SIZE`: The default page size for tables. This is the number of rows which are shown per page. Set


jbtronics:
The ability to download database dumps is quite critical, as it circumvents bascially any protection mechanism of Part-DB and allows access to even normally hidden infos like password hashes, secrets, etc.

Thats why there should be an environment variable to enable/disable the backup download ability.

Also this makes the downloadBackup endpoint quite critical. It should at very least have an CSFR token check. 
Maybe make the backup download the response of an POST form submit, where the user has to input his current password (because that is probably the most hard to forge).

Also all more critical endpoints, should also do an `$this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');`, to ensure that the user is not on some old remember me token.

Sebbeben:
Thanks for the security review @jbtronics! All suggestions have been implemented:

- ✅ `DISABLE_BACKUP_DOWNLOAD` env var (default `1` = disabled)
- ✅ Backup download via POST with CSRF token validation
- ✅ Password confirmation required before downloading
- ✅ `IS_AUTHENTICATED_FULLY` on all critical endpoints (createBackup, deleteBackup, deleteLog, downloadBackup, startUpdate, restore)

**Screenshot of the password confirmation dialog:**

Screenshot 2026-03-06 074855


Tested on a live instance — download works with password verification, wrong passwords are rejected, and the download button is hidden when `DISABLE_BACKUP_DOWNLOAD=1` (default).

jbtronics:
thanks

Issue #1256, opened 2026-02-19
Show part status in BOM


mkne:
To have an overview of parts in a BOM that are maybe NRND or even obsolete. It would be great to have the ability to show the column "Status" here:
<img width="1022" height="406" alt="Image" src="https://github.com/user-attachments/assets/bd191e87-3358-4d22-8dfe-5a2df0668803" />

like it can be done here:
<img width="1043" height="268" alt="Image" src="https://github.com/user-attachments/assets/933de221-afcc-4567-9a49-40c6ce10191e" />

Also having the average price (see #603 and #1221) and the new custom part state available as optional columns might be useful. 

This might help in cases when you have to do a redesign or to have an overview if parts are defined obsolete by the manufacturer. Also for calculation of last time buy order amounts it might be useful.

PR #1257, opened 2026-02-19
Show Part status in BOM

mkne:
Fixes Part-DB/Part-DB-server#1256


mkne:
This doesn't work on my test system. Probably needs to be added somewhere else.

PR #1260, opened 2026-02-19
Add KiCad HTTP Library API v2 with volatile field support

Sebbeben:
## Summary

Adds KiCad HTTP Library API v2 support with volatile field support, split out from #1241 as discussed.

The v2 spec introduces **volatile fields** — fields that are shown in KiCad's UI but NOT saved to the schematic. This is ideal for dynamic data like stock quantity and storage location.

### Changes

| File | What changed |
|------|-------------|
| `KiCadApiV2Controller.php` | New controller at `/kicad-api/v2/` with root, categories, category parts, and part detail endpoints |
| `KiCadHelper.php` | Added `int $apiVersion` parameter to `getKiCADPart()`, `$volatile` parameter to `createField()` |
| `KiCadApiV2ControllerTest.php` | Full test coverage: root endpoint, categories, parts, volatile fields, v1 vs v2 comparison, caching, auth |

### V2 API differences from V1

- **Root endpoint** returns links to categories (per v2 spec requirement)
- **Stock** field marked as `volatile: True` (shown in KiCad, not saved to schematic)
- **Storage Location** field marked as `volatile: True`
- All other fields remain non-volatile (saved to schematic as in v1)

### V2 spec reference

- Draft: https://gitlab.com/RosyDev/kicad-dev-docs/-/blob/http-lib-v2/content/apis-and-binding/http-libraries/http-lib-v2-00.adoc
- V1 (current): https://dev-docs.kicad.org/en/apis-and-binding/http-libraries/index.html

### Note

This PR is intentionally minimal — only the v2-specific additions. The underlying KiCad improvements (batch EDA editing, parameter export, populate command, etc.) were merged in #1241. This can be merged once the v2 spec is finalized, or kept as a draft until then.

### DIFF
--- a/src/Controller/KiCadApiV2Controller.php
+++ b/src/Controller/KiCadApiV2Controller.php
@@ -1,0 +1,107 @@
+<?php
+/*
+ * This file is part of Part-DB (https://github.com/Part-DB/Part-DB-symfony).
+ *
+ *  Copyright (C) 2019 - 2023 Jan Böhmer (https://github.com/jbtronics)
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU Affero General Public License as published
+ *  by the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+declare(strict_types=1);
+
+namespace App\Controller;
+
+use App\Entity\Parts\Category;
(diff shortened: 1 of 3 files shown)


jbtronics:
Are the new HTTP library things actual part of any KiCad release? I could not find anything about it for the upcoming KiCad 10.

In the linked documentation it seems that v2 was renamed to v1.1, and also includes other things than just a volatile property field (like timestamping)

Sebbeben:
Sorry for the late reply! I got a bit ahead of myself with this one — I assumed that since the v2 spec documentation existed, it would ship with KiCad 10, but that doesn't appear to be the case.

You're right on both points. The spec was renamed from v2 to v1.1, and it's much more than just volatile fields. After going through the full spec, v1.1 is primarily about **timestamp-based incremental caching** — the update endpoint, restructured parts/categories responses with timestamps and pagination, field type declarations (essential/detail/hidden), shorthand keys, etc. The volatile fields part is actually commented out throughout the spec and explicitly marked as "not currently supported by KiCad."

This PR only addressed the volatile fields, which is ironically the one part that isn't even active in the spec. I'd suggest closing this for now and revisiting once v1.1 actually lands in a KiCad release — at which point a proper implementation covering the full spec (especially the caching/timestamping) would make much more sense.

PR #1266, opened 2026-02-22
feat(parts table): add eda reference prefix and value columns

hrueger:
Hi,
I hope this makes sense 👍 I don't have my part descriptions formatted the same way everywhere, so I'd like to show this in a separate column.

Feel free to suggest any changes / improvements!


hrueger:
I'm unsure about what copilot suggests here, as none of the other `add` calls here do that. I have to confess though, that I don't fully understand the whole structure ;-)

jbtronics:
Thanks

PR #1280, opened 2026-03-03
/api/part_lots: add user_barcode filter

MayNiklas:
This pull request adds support for filtering `PartLot` entities by the `user_barcode` field via the API.
This is helpful, when interacting with the API using custom tooling.


MayNiklas:
Using LIKE filtering makes sense, since for a lot of suppliers, inner and outer bags have QR codes that differ in some small details.
That allows to treat them as the same stock in custom tooling.

I'm currently working on software that runs directly on scanners, allowing to interact with the Part-DB API, for moving lots around, adding lots, etc... 
Therefore being able to request a lot that matches a user_barcode is very helpful to me!
Will open source it soon for people interested.

jbtronics:
looks good. Thanks.

PR #1288, opened 2026-03-05
Fix: Reset PHP OPcache after update to prevent 500 errors

Sebbeben:
## Problem

After the update manager completes an update (git checkout, composer install, cache clear/warmup), PHP-FPM may still serve **stale bytecode from OPcache**. This causes 500 errors when the progress page auto-refreshes, because the old cached container tries to instantiate classes with changed constructor signatures.

Example error seen after updating from v2.7.1 to v2.8.1:
```
AttachmentURLGenerator::__construct(), 6 passed... and exactly 7 expected
```

This affects any native git installation running PHP-FPM with OPcache enabled (which is the default and recommended configuration).

## Solution

Added an OPcache reset step after cache warmup in the update, restore, and rollback flows. The approach:

1. Creates a temporary PHP script with a random token filename in `public/` (e.g., `_opcache_reset_<random>.php`)
2. Invokes it via HTTP (`http://127.0.0.1/` then `http://localhost/`) to trigger `opcache_reset()` in the PHP-FPM context
3. The script self-deletes after execution, with a safety cleanup in the `finally` block
4. If HTTP invocation fails (e.g., non-standard web server setup), it logs a warning and continues gracefully

### Why this approach?

- `opcache_reset()` from CLI only affects the CLI process, not PHP-FPM
- Restarting PHP-FPM requires root/sudo privileges
- External tools like `cachetool` may not be installed
- The temporary HTTP script approach is the most portable solution

## Changes

- `src/Services/System/UpdateExecutor.php`:
  - Added `resetOpcache()` private method
  - Added OPcache reset as Step 13 in the update flow (between cache warmup and maintenance mode off)
  - Added OPcache reset in the restore flow
  - Added OPcache reset after rollback
  - Updated total step counts accordingly

### DIFF
--- a/src/Services/System/UpdateExecutor.php
+++ b/src/Services/System/UpdateExecutor.php
@@ -208,6 +208,79 @@ class UpdateExecutor
     }
 
     /**
+     * Reset PHP OPcache for the web server process.
+     *
+     * OPcache in PHP-FPM is separate from CLI. After updating code files,
+     * PHP-FPM may still serve stale cached bytecode, causing constructor
+     * mismatches and 500 errors. This method creates a temporary PHP script
+     * in the public directory, invokes it via HTTP to reset OPcache in the
+     * web server context, then removes the script.
+     *
+     * @return bool Whether OPcache was successfully reset
+     */
+    private function resetOpcache(): bool
+    {
+        $token = bin2hex(random_bytes(16));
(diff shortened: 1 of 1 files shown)


jbtronics:
It is only required if `opcache.validate_timestamps=0`.

But i thinks its a good addition to include that. Thanks.

PR #1289, opened 2026-03-05
Show ManufacturingStatus in BOM

### LINKED ISSUE #1256: Show part status in BOM

mkne:
Fixes #1256
image

### DIFF
--- a/translations/messages.en.xlf
+++ b/translations/messages.en.xlf
@@ -12755,17 +12755,16 @@ Buerklin-API Authentication server:
         <target>Create [part] from barcode</target>
       </segment>
     </unit>
-<<<<<<< HEAD
     <unit id="xH258F." name="parts.create_from_scan.title">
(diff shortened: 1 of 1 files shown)


mkne:
@jbtronics Hi Jan, please merge this pull-request, if you consider it useful. I could finally fix it, see screenshot as proof. I don't understand, why Github shows that many commits but only the one file I changed. I closed the old PR because of this but now it's the same.

jbtronics:
thanks.

PR #1295, opened 2026-03-07
fix(scanner): use ZXing WASM detector for harder DigiKey Data Matrix labels

### LINKED ISSUE #1281: DigiKey Barcode Scanning does not work sometimes
First of all, thank you for implementing the feature to add parts by scanning the DigiKey Data Matrix. This will greatly help us add our existing stock to PartDB.

Nevertheless, we encountered some difficulties when scanning different DigiKey Data Matrix codes.

Description of the problem:
We scanned various DigiKey part bags with attached labels and Data Matrix codes. DigiKey ships parts in different types of bags, such as plain white plastic bags that need to be ripped open, with the Data Matrix printed directly onto the plastic; transparent plastic bags with a blue tint and a glued paper label containing the Data Matrix; and large plain white plastic bags used for reel shipments.

When scanning the transparent plastic bags, the Data Matrix is recognized and everything works fine. When scanning the large plain white bags for reel shipments, everything works fine as well. Unfortunately, when scanning the smaller plain white bags, the Data Matrix is not recognized. We believe that the Data Matrix codes on the smaller plain white bags are not recognized because they appear to be printed more boldly than on the other bags.

Tried solutions:
We have already tested different lighting conditions and ensured that the Data Matrix is not obscured by any reflections. Additionally, we tested with an iPhone 12 and a Google Pixel 10.

Furthermore, we tried scanning the different Data Matrix codes with other decoders and encountered the same problem: the codes printed on the plain white plastic bags were not recognized. It seems that many decoders have difficulty decoding these specific printed Data Matrix codes. However, the Data Matrix scanner provided by Inlite Research successfully decodes the problematic barcodes. This decoder is available at: https://online-barcode-reader.inliteresearch.com/

The following data matrices work:

Image
Image
Image

The following do not work (but are successfully decoded with the Inlite decoder):

Image
Image
Image

SergioChan:
## Summary
- force Part-DB's scanner path to use the ZXing WASM BarcodeDetector ponyfill instead of the browser-native detector
- keep the existing local WASM override wiring, but make the detector choice deterministic for scanning
- this targets issue reports where native detector paths fail on bolder/low-contrast DigiKey Data Matrix prints

## Testing
- not runnable in this checkout: frontend install fails because required Symfony UX asset paths under vendor/ are absent (vendor/symfony/ux-translator/assets), so yarn install --frozen-lockfile cannot complete
- manual verification plan documented in PR: confirm previously failing DigiKey white-bag labels decode on Chrome/Android + iOS after rebuilding assets

## Related
Fixes #1281

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -48,6 +48,7 @@ global.$ = global.jQuery = require("jquery");
 
 //Use the local WASM file for the ZXing library
 import {
+    BarcodeDetector as ZXingBarcodeDetector,
     setZXingModuleOverrides,
 } from "barcode-detector/ponyfill";
 import  wasmFile from "../../node_modules/zxing-wasm/dist/reader/zxing_reader.wasm";
@@ -59,3 +60,7 @@ setZXingModuleOverrides({
         return prefix + path;
     },
 });
+
+// Prefer the ZXing WASM detector over inconsistent native BarcodeDetector implementations.
+// This improves Data Matrix decoding reliability for difficult DigiKey labels (see #1281).
+globalThis.BarcodeDetector = ZXingBarcodeDetector;


jbtronics:
The library already uses the polyfill as fallback, when the native barcode detector does not give an result, so I would not expect any improvements with that PR.

And the choice of trying the native detector first was deliberate, as in my experiments it was faster and more reliable than zxing-wasm for many barcodes (which are not datamatrix)

SergioChan:
Thanks for the detailed context — that makes sense.

Given the current flow already falls back to the polyfill, I agree this patch probably isn’t a net improvement as-is. I can close this PR unless you want a narrower follow-up focused on measurable behavior (for example, adding a regression test with a specific hard Data Matrix sample where fallback selection is verifiable).

jbtronics:
I would close it for now. If you have some data showing a siginificant improvement we can discuss how to move forward.

Issue #1297, opened 2026-03-08
Discussion: unsure if MPN fallback is a good idea


MayNiklas:
Hello!

First, I want to say that I love the improvements Part-DB made over the last month!
A lot of features I wished for over the last year got merged, making a lot of features more helpful!

I've extensively used the LCSC scan feature over the last week.
I'm unsure if the current implementation of falling back to the MPN is a good idea:

https://github.com/Part-DB/Part-DB-server/blob/343c078b7d0c416a34235c930e0d8d147e06fd08/src/Services/LabelSystem/BarcodeScanner/BarcodeScanResultHandler.php#L207-L214

For example:
I do have 6 different versions of the AMS1117-3.3 from different manufacturers.
They have VERY different specs regarding their supported input voltages.

Currently, when Part-DB only has one of them added as a part, scanning the rest would lead to the first part.

I had this issue last week, which really confused me.

In my opinion, there are 2 possible solutions:

1. At least add a warning when being redirected to a part with a different SPN ("hey, we used the MPN fallback, be careful")
2. Removing the fallback and treating the part as a different one, because its SPN does not match.

Currently, I always have to be really careful, not adding a stock to the wrong part, since a lot of my parts share their MPN with each other.

What are your thoughts? @swdee @jbtronics 
I'm very thankful for the new features, so please don't get me wrong!
I just want to question, if the MPN fallback should stay as it is currently.
It currently requires some extra caution, which might be difficult in multi user enviroments.

swdee:
The fallback code you highlighted is triggered for EIGP114 type barcodes (Digikey, Mouser).

https://github.com/Part-DB/Part-DB-server/blob/343c078b7d0c416a34235c930e0d8d147e06fd08/src/Services/LabelSystem/BarcodeScanner/BarcodeScanResultHandler.php#L130-L154

There is separate fallback for LCSC barcodes here;

https://github.com/Part-DB/Part-DB-server/blob/343c078b7d0c416a34235c930e0d8d147e06fd08/src/Services/LabelSystem/BarcodeScanner/BarcodeScanResultHandler.php#L223-L241

However as LCSC barcodes (to my knowledge) always have the PC code (LCSC Code that begins with letter C) I would not expect the code to hit the MPM fallback.

Would you provide the raw barcode scan data for the multiple `AMS1117-3.3` parts you have so we can take a closer look?

MayNiklas:
@swdee you are right about me citing the wrong code snipped!

```
[...]

PR #1302, opened 2026-03-08
Bug fix: Remove fallback from LCSC barcode part resolver

swdee:
As discussed in #1297 this PR removes the fallback logic from the part resolver of LCSC codes.   This resolves the issue where a similarly named manufacture part but with different LCSC code `pc` number was being matched when it should not have been.

### DIFF
--- a/src/Services/LabelSystem/BarcodeScanner/BarcodeScanResultHandler.php
+++ b/src/Services/LabelSystem/BarcodeScanner/BarcodeScanResultHandler.php
@@ -142,8 +142,7 @@ final readonly class BarcodeScanResultHandler
         }
 
         if ($barcodeScan instanceof LCSCBarcodeScanResult) {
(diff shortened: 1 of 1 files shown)


jbtronics:
I think it is useful to have a fallback using PartRepository::getPartBySPN with the lcsc part number.
That way the resolving is also useable if the part was not added via the LCSC provider

swdee:
> I think it is useful to have a fallback using PartRepository::getPartBySPN with the lcsc part number. That way the resolving is also useable if the part was not added via the LCSC provider

This has been added in.

jbtronics:
The SPN resolval logic should also be part of the LCSC submethod (or we inline the complete method).

And I think most people will have the LCSC product code as SPN (as that is the order number) not with the MPN.

swdee:
Ok, I think the changes I have made address your points.

jbtronics:
thanks

PR #1316, opened 2026-03-12
Add Quick Apply and batch update to bulk info provider import

Sebbeben:
## Summary

Adds one-click **Quick Apply** buttons and a batch **Apply All (Top Results)** action to the bulk info provider import step 2 page, so users no longer need to open every single component and click save twice to update from provider data.

This was requested by a user: *"For something that should be added is on bulk import to be able to just mass update from provider rather than having to open EVERY single component and click save changes twice, so annoying"*

### Features
- **Quick Apply per result**: AJAX button that fetches provider details and merges them into the part in one click
- **Apply All (Top Results)**: Batch button that applies the top-priority result for every pending part at once
- **Match highlighting**: Green "Match" badge when the source keyword matches the result (MPN, SPN, name)
- **Top result indicator**: First result (highest priority) is highlighted with a green row and "Top" badge
- **Active/History split on manage page**: Active jobs shown at top, completed/stopped jobs in a separate History section
- **Job ID display**: Each job now shows `#ID` for easy identification
- **Navigation buttons**: "Back to Jobs" and "Back to Parts" on the step 2 review page
- **Completion banner**: Clear indication when a job is fully completed
- **Auto-priority for field mappings**: Each new field mapping row auto-increments priority
- **Stale job cleanup**: Pending jobs with 0 results are automatically cleaned up

### Bug fixes
- Fix TypeError in LCSCProvider when keyword is numeric string (cast to string)
- Fix 500 error when field mapping has null field or no search results
- Fix spinning icon on manage page (was using pulse animation that never stopped)
- Fix text visibility issues in dark theme for source field badges
- Remove unexpected "reason for skipping" prompt when clicking Quick Apply

### Screenshots

**Step 2 - Review page with Quick Apply and match highlighting:**
Screenshot 2026-03-12 115444

**Manage page with Active/History split and job IDs:**
Screenshot 2026-03-12 115518

## Test plan
- [x] PHPStan: 0 errors on all changed files
- [x] PHPUnit: All new tests pass (8 new test methods, 16+ assertions)
- [x] Manual testing on live server with real LCSC provider data
- [x] Quick Apply on individual parts with single and multiple results
- [x] Apply All (Top Results) batch operation
- [x] Match highlighting for MPN, SPN, and name matches
- [x] Navigation buttons and completion flow
- [x] Active/History job separation on manage page

### DIFF
--- a/assets/controllers/bulk_import_controller.js
+++ b/assets/controllers/bulk_import_controller.js
@@ -3,14 +3,16 @@ import { generateCsrfHeaders } from "./csrf_protection_controller"
 
 export default class extends Controller {
     static targets = ["progressBar", "progressText"]
-    static values = { 
+    static values = {
         jobId: Number,
         partId: Number,
         researchUrl: String,
         researchAllUrl: String,
         markCompletedUrl: String,
         markSkippedUrl: String,
-        markPendingUrl: String
+        markPendingUrl: String,
+        quickApplyUrl: String,
+        quickApplyAllUrl: String
     }
 
     connect() {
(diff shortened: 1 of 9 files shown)


jbtronics:
This PR seems to contains the commits of PR #1330 
Would be good to have it separated. Or i wait until the other PR is merged, then it should be easier to diff.

Sebbeben:
Good catch — the stray Watchtower commit has been dropped via force-push. The branch now contains only the 14 bulk-import commits, so the diff should be clean to review.

All CI checks (PHPUnit across PHP 8.2–8.5 × mysql/postgres/sqlite, static analysis, build) are green on the updated branch.

jbtronics:
thanks.
I fixed some things related to when new datastructures would be required.
It shows some more user friendly error message, and at least manufacturers and suppliers are created automatically, as there is potentially a wide variety.

Sebbeben:
Thank you for that, and sorry for my late reply. I have been busy

PR #1330, opened 2026-03-25
Add Docker update support via Watchtower integration

Sebbeben:
## Summary

Extends the Update Manager to support Docker-based installations by integrating with [Watchtower](https://github.com/containrrr/watchtower), a popular container update tool.

- **WatchtowerClient service** — communicates with Watchtower's HTTP API to check status and trigger updates
- **Docker update progress page** — Stimulus-based polling UI that survives container restarts, with progress indicators and timeout handling
- **Health endpoint** (`/update-manager/health`) — unauthenticated JSON endpoint for post-restart detection (session is lost when container restarts)
- **Docker-specific UI** — hides Git-only fields (branch, commit, local changes) for Docker installs, shows Watchtower connection status
- **Pre-update backup** — creates database backup before triggering Watchtower
- **CSP fix** — replaces `javascript:history.back()` with proper navigation in error template

### How it works

1. User clicks "Update to latest version" on the Update Manager page
2. Part-DB creates a database backup
3. Part-DB calls Watchtower's HTTP API to trigger an update
4. User is redirected to a progress page that polls the health endpoint
5. When the container restarts with the new image, the health endpoint responds and the UI shows success

### Configuration

Docker users add a Watchtower sidecar to their `docker-compose.yml`:

```yaml
watchtower:
  image: containrrr/watchtower
  volumes:
    - /var/run/docker.sock:/var/run/docker.sock
  environment:
    - WATCHTOWER_HTTP_API_UPDATE=true
    - WATCHTOWER_HTTP_API_TOKEN=your-secret-token
    - WATCHTOWER_LABEL_ENABLE=true
  ports:
    - "8080:8080"
```

And set environment variables for Part-DB:
```
WATCHTOWER_API_URL=http://watchtower:8080
WATCHTOWER_API_TOKEN=your-secret-token
```

If Watchtower is not configured, the Update Manager gracefully shows an informational message explaining how to enable Docker updates.

### Files changed

| File | Purpose |
|------|---------|
| `src/Services/System/WatchtowerClient.php` | New — Watchtower HTTP API client |
| `src/Controller/UpdateManagerController.php` | Docker update actions + health endpoint |
| `src/Services/System/UpdateChecker.php` | Watchtower status in update info |
| `src/Services/System/UpdateExecutor.php` | Docker update validation |
| `src/Services/System/InstallationType.php` | Docker supports auto-update |
| `src/EventSubscriber/MaintenanceModeSubscriber.php` | Whitelist health endpoint |
| `assets/controllers/docker_update_progress_controller.js` | New — Stimulus polling controller |
| `templates/admin/update_manager/docker_progress.html.twig` | New — Progress page template |
| `templates/admin/update_manager/index.html.twig` | Docker-specific UI sections |
| `templates/bundles/TwigBundle/Exception/error.html.twig` | CSP fix for back link |
| `translations/messages.en.xlf` | Translation keys for Docker update UI |

## Test plan

- [ ] PHPStan passes (verified, 0 errors)
- [ ] Tested on Docker installation (Orange Pi ARM64) with Watchtower sidecar
- [ ] Verified update progress UI survives container restart
- [ ] Verified graceful degradation when Watchtower is not installed
- [ ] Verified health endpoint is accessible without authentication
- [ ] Verified CSRF protection on all mutating endpoints

### Update Manager - Docker Installation
update_manager_docker

### Update in Progress
updating

### Update Complete
update_complete
note: this is when i was developing and thats why it says updating from v2.9.0 -> v2.9.0

### DIFF
--- a/.env
+++ b/.env
@@ -76,6 +76,12 @@ DISABLE_BACKUP_RESTORE=1
 # When enabled, users must confirm their password before downloading.
 DISABLE_BACKUP_DOWNLOAD=1
 
+# Watchtower integration for Docker-based updates.
+# Set these to enable one-click updates via the Update Manager UI.
+# See https://containrrr.dev/watchtower/ for Watchtower setup.
+WATCHTOWER_API_URL=
+WATCHTOWER_API_TOKEN=
+
 ###################################################################################
 # SAML Single sign on-settings
 ###################################################################################
--- a/assets/controllers/docker_update_progress_controller.js
+++ b/assets/controllers/docker_update_progress_controller.js
@@ -1,0 +1,377 @@
+/*
+ * This file is part of Part-DB (https://github.com/Part-DB/Part-DB-symfony).
+ *
+ *  Copyright (C) 2019 - 2026 Jan Böhmer (https://github.com/jbtronics)
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU Affero General Public License as published
+ *  by the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import { Controller } from '@hotwired/stimulus';
+
+/**
+ * Stimulus controller for Docker update progress tracking.
+ *
+ * Polls the health check endpoint to detect when the container restarts
+ * after a Watchtower-triggered update. Drives the step timeline UI
+ * with timestamps, matching the git update progress style.
+ */
+export default class extends Controller {
+    static values = {
+        healthUrl: String,
+        previousVersion: { type: String, default: 'unknown' },
+        pollInterval: { type: Number, default: 5000 },
+        maxWaitTime: { type: Number, default: 600000 }, // 10 minutes
+        // Translated UI strings (passed from Twig template)
+        textPulling: { type: String, default: 'Waiting for Watchtower to pull the new image...' },
+        textPullingDetail: { type: String, default: 'Watchtower is checking for and downloading the latest Docker image...' },
+        textRestarting: { type: String, default: 'Container is restarting with the new image...' },
+        textRestartingDetail: { type: String, default: 'The container is being recreated with the updated image. This may take a moment...' },
+        textSuccess: { type: String, default: 'Update Complete!' },
+        textSuccessDetail: { type: String, default: 'Part-DB has been updated successfully via Docker.' },
+        textTimeout: { type: String, default: 'Update Taking Longer Than Expected' },
+        textTimeoutDetail: { type: String, default: 'The update may still be in progress. Check your Docker logs for details.' },
+        textStepPull: { type: String, default: 'Pull Image' },
+        textStepRestart: { type: String, default: 'Restart Container' },
+    };
+
+    static targets = [
+        // Header
+        'headerWhale', 'titleIcon',
+        'statusText', 'statusSubtext',
+        'progressBar', 'elapsedTime',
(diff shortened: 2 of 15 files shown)


jbtronics:
I think it would be good to have something in the docker installation documentation on how to use setup that.

Sebbeben:
> I think it would be good to have something in the docker installation documentation on how to use setup that.

Added Watchtower setup documentation to the Docker installation guide (`docs/installation/installation_docker.md`). It covers:

- Adding the Watchtower service to `docker-compose.yaml`
- Configuring `WATCHTOWER_API_URL` and `WATCHTOWER_API_TOKEN` environment variables
- Using label-based filtering to only manage the Part-DB container
- How the Update Manager UI integrates with Watchtower

Also squashed all commits into a single clean commit and addressed the PHPStan issue (optional parameter before required ones).

Note: The postgres CI failures are a pre-existing upstream issue (FK violation during fixture loading) — they also fail on master.

Sebbeben:
### Updated Docker Update Progress UI

The progress page has been redesigned to match the git update style:

[Docker Update Progress]new_update_screen




**Changes in this update:**
- Animated Docker Moby whale logo (official SVG, Docker blue #2496ED)
- 6 detailed steps with live timestamps: Trigger → Pull → Stop → Restart → Health Check → Verify
- CSP-compatible progress bar using CSS classes instead of inline styles
- Translated UI strings passed from Twig to Stimulus controller
- WatchtowerClient made nullable for non-Docker installations (fixes crash on git installs)
- Watchtower setup documentation added to `docs/installation/installation_docker.md`
- All commits squashed into one clean commit

jbtronics:
I am not sure how wise it is, to build this around watchtower.
According to the github project README it is not maintained anymore: https://github.com/containrrr/watchtower

And the last release is from 2023.

Sebbeben:
That's a fair concern — I should have checked the maintenance status more carefully. Here are the options I see for keeping the one-click update UX this PR aims for:

**Option 1: Switch to a maintained fork (smallest change)**
[`nicholas-fedor/watchtower`](https://github.com/nicholas-fedor/watchtower) is an active fork — v1.16.1 released April 2026, same HTTP API (`POST /v1/update` with bearer token). Drop-in swap: zero code changes in Part-DB, just update the documented image reference in `docs/installation/installation_docker.md`. Tradeoff: still a single-maintainer fork, so bus factor is low.

**Option 2: Abstract the backend behind an interface**
Introduce a `ContainerUpdaterInterface` with Watchtower as one implementation. Users configure `CONTAINER_UPDATER=watchtower|…` and we can add new backends as the ecosystem evolves. Future-proofs the feature if forks also stagnate. More code to maintain, but keeps options open.

**Option 3: DIY via `docker.sock`**
Part-DB talks to the Docker Engine API directly (pull image, recreate container via a detached helper, since a container can't cleanly restart itself mid-request). No third-party dependency, but mounting `docker.sock` in a web app container is a meaningful security surface — effectively root on the host. Would need [`tecnativa/docker-socket-proxy`](https://github.com/Tecnativa/docker-socket-proxy) with a minimal allow-list to mitigate. More complex, and Docker Engine API doesn't understand compose so reconstructing run args is fiddly.

My lean is **Option 1** (swap to `nicholas-fedor/watchtower`) for this PR — minimal change, we ship the feature now — and if a second viable backend appears later we can refactor to Option 2 at that point. Option 3 only makes sense if we want to eliminate the sidecar entirely, and the security tradeoff feels heavy for the benefit.

What do you think?

jbtronics:
Yeah using the fork is probably the best solution for now

Sebbeben:
Done — pushed `43b9441b` switching the Docker installation docs to [`ghcr.io/nicholas-fedor/watchtower:latest`](https://github.com/nicholas-fedor/watchtower) and added a short info note explaining why (original unmaintained since Nov 2023, fork is drop-in API-compatible).

No code changes — `WatchtowerClient` talks the same HTTP API to either image.

d-buchmann:
Yesterday heise [featured](https://www.heise.de/ratgeber/Watchtower-und-Alternativen-So-halten-Sie-Docker-Container-automatisch-aktuell-11243856.html) watchtower alternatives - one being Whats up Docker (WUD).
Just in case you happen to implement Option 2 someday, that should probably go onto that list =)

jbtronics:
I finally got to reviewing the PR, looks good.
Thank you.

PR #1337, opened 2026-04-04
Fix creating parts from TME if the SPN contains percent signs

alufers:
Example: https://www.tme.eu/en/details/smd0603-5k1-1%25/smd-resistors/royalohm/0603saf5101t5e/ 

The TME API returns an URL for that part which contains an unencoded percent sign, this later causes a form validation error when trying to save the added part.

image

Fixed by percent encoding it if detected. Also added unit tests for the TMEProvider.


jbtronics:
Thanks. Looks good.
The test is also very nice.

PR #1338, opened 2026-04-04
Fix #1305: Enable BOM sorting on part fields (Storage location, Manufacturing status) and fix BOM table query/pagination issues

mkne:
This PR refactors ProjectBomEntriesDataTable to use the same robust two-step query strategy as PartsDataTable, resolving multiple query/pagination errors on the BOM page and restoring expected column behavior.

**What changed**

- Switched BOM datatable adapter to TwoStepORMAdapter.
- Added an ID-only filter_query for filtering/sorting/pagination.
- Added a detail_query for fetching full BOM entities and joins.
- Kept stable row ordering between filter/detail queries via FieldHelper::addOrderByFieldParam(...).
- Preserved storage location sorting support with aggregate ordering (NATSORT(MIN(_storelocations.name))).
- Added explicit sorting for Manufacturing Status (orderField => 'part.manufacturing_status').
- Aligned datatable entity usage with ProjectBOMEntry.

**Why**
The previous single-query fetch-join approach caused Doctrine/PgSQL failures such as:

- non-unique scalar/count query results,
- grouping errors,
- and Iterate with fetch join ... PartLot ... not allowed.
Using TwoStepORMAdapter avoids these fetch-join pagination pitfalls and keeps sorting/filtering functional.

**User-visible impact**
- BOM page loads reliably (including repeated loads/callbacks).
- Manufacturing Status column now displays values again.
- Manufacturing Status column is now sortable.

**Testing performed**

- Verified BOM datatable code path compiles (php -l)
- Manually validated query setup for normal BOM load
- tested storage location sorting
- tested manufacturing status display and sorting

Disclaimer: This description was written by Codex but manually edited and corrected.

### DIFF
--- a/src/DataTables/ProjectBomEntriesDataTable.php
+++ b/src/DataTables/ProjectBomEntriesDataTable.php
@@ -152,7 +152,6 @@ class ProjectBomEntriesDataTable implements DataTableTypeInterface
 
             ->add('manufacturing_status', EnumColumn::class, [
                 'label' => $this->translator->trans('part.table.manufacturingStatus'),
-		        'data' => static fn(ProjectBOMEntry $context): ?ManufacturingStatus => $context->getPart()?->getManufacturingStatus(),
                	'class' => ManufacturingStatus::class,
                 'render' => function (?ManufacturingStatus $status, ProjectBOMEntry $context): string {
                     if ($status === null) {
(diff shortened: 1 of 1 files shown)


jbtronics:
Looks good. Thanks

PR #1342, opened 2026-04-10
Add custom KiCad autocomplete list settings

Dant-hw:
## Summary

  Add admin settings for KiCad autocomplete lists.

  This change keeps the autogenerated default KiCad lists read-only and introduces editable custom override files:

  - `public/kicad/footprints_custom.txt`
  - `public/kicad/symbols_custom.txt`

  A new setting controls whether KiCad autocomplete uses the autogenerated defaults or the custom files.

  ## What changed

  - Added a `useCustomList` KiCad setting
  - Updated KiCad autocomplete fields to use either the default or custom list files
  - Updated the admin editor page to:
    - show custom files as editable
    - show default files as read-only reference
  - Added seeded custom files with a comment header
  - Updated docs and translations
  - Extended the controller test coverage

  ## Behavior

  - Default KiCad lists remain autogenerated and unchanged
  - Custom lists are stored separately and can be edited in the UI
  - When `Use custom autocomplete lists` is enabled, autocomplete uses the custom files instead of the default files
  - This is a full switch, not a merge of default and custom entries

  ## Why this approach

  This avoids editing autogenerated files directly and keeps user-maintained autocomplete entries stable across updates, while keeping the implementation small
  and localized.

  ## Verification

  - Verified locally in Docker
  - Ran the focused PHPUnit test:
    - `tests/Controller/KicadListEditorControllerTest.php`


jbtronics:
Looks good. thank you.

I had to change how the custom files are generated. Because if they are tracked via git, then it breaks the update manager and its easy to accidentially ignore them.  They are now ignored by git and created on the fly on cache warmup, or if the manager tries to read them.

Also the autocomplete lists are now included in the config backups created by the backup command.

Dant-hw:
Thank you so much!

PR #1344, opened 2026-04-15
Add SI-prefix-aware sorting column for parts tableFeature/si value sort

wschopohl:
## Summary

  - Adds an optional **"SI Value"** column to the parts table that parses numeric values with SI prefixes (p, n, u/µ, m, k/K, M, G, T) from part names and sorts by the resulting physical value
  - Useful for electronic components where alphabetical sorting produces wrong results — e.g. `100nF`, `10pF`, `1uF` sort as `10pF < 100nF < 1uF`
  - The column displays the parsed float value, or an empty cell for non-matching names
  - Only matches numbers at the start of the name — part numbers like "MCP2515" or "Crystal 20MHz" are not affected
  - Supports both dot and comma decimal separators (European locale)
  - Falls back to NATSORT for parts without SI values, so they sort alphabetically rather than arbitrarily
  - Opt-in: not in default columns, users enable it via the column visibility button

  ### Implementation

  - New `SiValueSort` DQL function with platform-specific SQL for PostgreSQL (POSIX regex), MySQL/MariaDB (REGEXP_SUBSTR), and SQLite (PHP callback via the existing middleware)
  - 33 unit tests covering all platforms, SI prefixes, edge cases, and sort order verification

  ## Test plan

  - [x] Unit tests pass (33 tests, all platforms)
  - [x] PHPStan level 2 passes
  - [x] Tested against real parts database with SQLite

### DIFF
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -56,6 +56,7 @@ doctrine:
                 natsort: App\Doctrine\Functions\Natsort
                 array_position: App\Doctrine\Functions\ArrayPosition
                 ilike: App\Doctrine\Functions\ILike
+                si_value_sort: App\Doctrine\Functions\SiValueSort
 
 when@test:
     doctrine:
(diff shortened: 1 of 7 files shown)


wschopohl:
SI Value Demo

For completeness, here is a screenshot showing how SI Value improves table ordering. It uses the value from 'Name' under the hood, extracting leading numbers with SI prefixes.

jbtronics:
Looks good. Thanks.
That should be quite useful.

wschopohl:
Awesome and thanks for fixing the PHPStan issue - must have slipped my attention

PR #1345, opened 2026-04-15
Add price columns to project BOM table and build price summary

MayNiklas:
This is a continuation of #1221, which went stale.

The implementation has been improved with the suggestions I made under the original PR and rebased onto the current master and several issues from the review have been addressed.
Thanks for @Whinis for the idea!

BOM table:
- Price: unit price for each BOM entry in the base currency
- Extended Price: unit price × BOM quantity

Project info tab:
- Shows the total BOM price for N builds as a badge on the info tab
- Adjustable via a small "number of builds" input; bulk pricing tiers are taken into account (ordering 10× a part may yield a lower per-unit price)
- When N > 1, the per-unit project price is shown alongside the total Implementation notes

- When a BOM entry's quantity falls below the part's minimum order amount, the minimum order amount is used for the price tier lookup (otherwise calculateAvgPrice returns null and the price would silently show as 0)
- Prices are rounded up to 2 decimal places (e.g. 0.001 € → 0.01 €) to avoid displaying 0.00 for very cheap parts
- All price rendering goes through MoneyFormatter for locale-aware formatting with the correct currency symbol

### DIFF
--- a/src/Controller/ProjectController.php
+++ b/src/Controller/ProjectController.php
@@ -69,10 +69,13 @@ class ProjectController extends AbstractController
             return $table->getResponse();
         }
 
+        $number_of_builds = max(1, $request->query->getInt('n', 1));
+
         return $this->render('projects/info/info.html.twig', [
             'buildHelper' => $buildHelper,
             'datatable' => $table,
             'project' => $project,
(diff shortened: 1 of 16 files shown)


jbtronics:
Looks good. Thank you.

PR #1346, opened 2026-04-16
Fix sort order after column reorder on page reload

wschopohl:
## Summary

- Fixes incorrect sort order when columns have been reordered via colReorder drag-and-drop and the page is reloaded
- The saved sort state uses visual (post-reorder) column indices, but `initial_order` was sending them to the server as-is — the server interpreted them as original indices, sorting the wrong column
- Uses the saved `colReorder` mapping to translate visual indices back to original indices before sending `initial_order` in the `_init` request

## Steps to reproduce

1. Open any DataTables view (e.g. parts list)
2. Drag a column to reorder it
3. Click a column header to sort
4. Reload the page
5. **Before fix:** wrong column is sorted. **After fix:** correct column is sorted.

### DIFF
--- a/assets/controllers/elements/datatables/datatables_controller.js
+++ b/assets/controllers/elements/datatables/datatables_controller.js
@@ -113,8 +113,16 @@ export default class extends Controller {
                     return null;
                 }
 
+                //The saved order index is visual (post-reorder). If colReorder state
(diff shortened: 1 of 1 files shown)


jbtronics:
Looks good.
Thanks.

PR #1352, opened 2026-04-24
Add 'Add stock' button to part stock info page

kernchen-brc:
### Summary
This PR adds an "Add stock" button to the part stock information page.

### Motivation
Previously, adding new stock entries required navigating away from the part stock
information view. This change improves usability by allowing users to create stock
entries directly from the relevant page.

### Changes
- Added an "Add stock" button to the part stock info page

### Result
- Fewer clicks to add stock

### DIFF
--- a/config/reference.php
+++ b/config/reference.php
@@ -1712,7 +1712,7 @@ use Symfony\Component\Config\Loader\ParamConfigurator as Param;
  *     length?: scalar|Param|null, // Default: 5
  *     width?: scalar|Param|null, // Default: 130
  *     height?: scalar|Param|null, // Default: 50
- *     font?: scalar|Param|null, // Default: "/home/jan/php/Part-DB-server/vendor/gregwar/captcha-bundle/DependencyInjection/../Generator/Font/captcha.ttf"
+ *     font?: scalar|Param|null, // Default: "/home/user/documents/Part-DB-server/vendor/gregwar/captcha-bundle/DependencyInjection/../Generator/Font/captcha.ttf"
  *     keep_value?: scalar|Param|null, // Default: false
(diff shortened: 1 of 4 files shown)


kernchen-brc:
The new commit adds a checkbox that allows disabling the warning banner for user-uploaded HTML attachments. The banner remains enabled by default.

kernchen-brc:
## Last commit in this pull request.  

Add a new system setting "Allow HTML attachments to use browser storage"
(disabled by default) that controls whether sandboxed HTML attachments
can access the browser's localStorage/sessionStorage via allow-same-origin.

When disabled (default), attachments run in an opaque origin as before,
preventing any persistent storage. When enabled, data survives page
reloads but grants the attachment access to Part-DB's browser origin.

jbtronics:
Can you split it up into two PRs? One for the sandbox things one for the storage managment.

I dont really like the allow-same-origin mechanism, as it bascially breaks the sandboxing completley. This allows HTML attachments to steal the users authentication and send it to a third party webpage. And afterwards they can do anything they want as the user. And all that just from opening an malicous HTML attachment just for a brief moment.

If at all, that option should only be configurable via environment variables, so that only the server administrator can do that and it does not happen randomly.

What is your usecase for it? Maybe there is a better solution.

kernchen-brc:
So I removed the commits related to the sandboxing options. This pull request now only contains the additional "Add stock" button.

The original use case for the sandbox changes was to support KiCad’s Interactive HTML BOM directly in the browser, without requiring users to download the attachment. Currently, it relies on the Web Storage API (e.g. localStorage) to persist data like placed components between sessions, which no longer works under strict sandboxing.
However, I agree that relaxing the sandbox (especially via allow-same-origin) is not a safe solution.

Do you have any suggestions on how this could be implemented more securely?

jbtronics:
Looks good, thanks.

PR #1380, opened 2026-05-22
Basic multi-build implementation.

davebranton:


sziszi90 (commented):
Thank you for this commit. It is very useful.  Please could you group the parts by the best price when multiple suppliers are available (right now, I think it only considers the first supplier)?

davebranton:
It does consider only the first supplier, if there are any. But at the moment, the lack of translations or tests would probably prevent this commit from making it into main anyway. 

I'm not really sure exactly what "group parts by the best price" means anyway? It could attempt to determine the cheapest supplier using the information about price breaks per supplier, but maybe it's best to just provide links to all the suppliers instead of trying to get too clever about it. Supplier's prices can change, and their stock might be out anyway.

As well, the code probably isn't structured the way the the maintainer of this repo likes, so I'm kind of waiting for some input from him.

PR #1404, opened 2026-06-08
Update hide_sidebar_controller.js

d-buchmann:
With this fix, the sidebar state is reapplied correctly on page reload.


jbtronics:
thanks.

Issue #1355, opened 2026-04-26
search, % as a wildcard works inconsistently


RaptorDE:
Since PR [#1075](https://github.com/Part-DB/Part-DB-server/issues/1075#issuecomment-3848754596), the % character is no longer consistently recognized as a wildcard. It works in the quick search, but no results are displayed in the full search.

This change makes the search function worse for power users. You can use Reg.Ex. as a workaround, but .* is not as convenient as % as a wildcard. I would appreciate it if there were an option in the settings to restore the old behavior.

<img width="426" height="239" alt="Image" src="https://github.com/user-attachments/assets/921680bc-487a-4fa0-8b31-3c7d4b018eb2" />

<img width="885" height="348" alt="Image" src="https://github.com/user-attachments/assets/bd6c443e-764a-4edf-ae78-348a66bc1900" />

jbtronics:
Implemented by PR #1406

PR #1406, opened 2026-06-08
Implement advanced search

d-buchmann:
Supersedes PR #944
Nota bene: Test were generated using AI. The actual code is handwritten.

Up to 5 individual tokens (separated by spaces) can be given as search string. Each token is individually searched for in all selected fields.

Examples (assuming the relevant fields are selected for search):
- a part named `foo` with a tag `bar` will be found with the search string "foo bar".
- a part named `bar baz` will be found with the search string "baz bar".
- A capacitor named `MLCC 100nF`, Manufacturer `Kemet`, Footprint `0603` will be found with any combination of the 4 keywords in any order.

All above examples don't yield these results in the current search implementation.
While the filter would be the more efficient tool to achieve the same results, some users might not be familiar with this - but quite possibly with a search bar.

Added SearchSettings which has 3 items:
- Enable advanced search (or else fall back to the current behavior)
- Search token count (default 3, range 2-5), limits the query size/complexity
- Configurable behavior of wildcards (#1355)


jbtronics:
I took a rough look:

* I think its not so useful to make this search mode global behavior, but instead user configurable (via the existing search options dropdown). This also allows to temporarily switch it. Maybe the system settings can stay as some default
* Is there much need to make the token count  configurable? Is there much disadvantage than just hardcode it to 5 or so? Its probably rare that someone use so many token anyway, and the worst that can happen is that the search request takes a bit longer, but you can still use a different search mode for this.

d-buchmann:
You're right about the user configurability, that way it can adapt to the varying user preferences much better.
I also hardcoded the token count to 5 which is indeed plenty and threw out the search settings completely because that would have overcomplicated it.

Meanwhile `advanced search` sounds odd to me, so I used the term `extensive` which describes better what it does - it just searches "more", not necessarily "better".

~~One more aspect I haven't yet solved is that regex and extensive search are mutually exclusive in the PHP code, but not in the user interface. It works, but if regex is enabled, the other two options will be silently ignored.~~

PR #1408, opened 2026-06-09
Parameter parser AWG

### LINKED ISSUE #1351: Min-max validation for AWG unit
Hello,

When importing components, such as wire terminals, which have the wire diameter parameter in AWG units then the value validation breaks because AWG units have their values in reverse (the smaller the diameter, the bigger the AWG value).

An example of such component might be the following:
[https://www.tme.eu/en/details/dg300-5.0-2p12/pcb-terminal-blocks/degson-electronics/dg300-5-0-02p-12-00a-h/](https://www.tme.eu/en/details/dg300-5.0-2p12/pcb-terminal-blocks/degson-electronics/dg300-5-0-02p-12-00a-h/)

The expected behaviour would be reversing the min and max values in the comparison for this specific unit.

**Server Side**
- Part-DB Version: 2.9.1 (Docker)
- PHP Version: 8.4.18 (SAPI: fpm-fcgi)
- Database Server: sqlite 3.40.1

d-buchmann:
Swaps the min/max AWG values if they are ordered from "small diameter" to "big diameter", i.e. "big number" to "small number".
Fixes #1351.

### DIFF
--- a/src/Services/InfoProviderSystem/DTOs/ParameterDTO.php
+++ b/src/Services/InfoProviderSystem/DTOs/ParameterDTO.php
@@ -94,6 +94,13 @@ readonly class ParameterDTO
 
                 //If both parts have the same unit and both values are numerical, we'll save it as range
                 if ($unit === $unit2 && is_numeric($number) && is_numeric($number2)) {
+                    //If the parameter contains AWG in name or unit and min/max order is "wrong", swap it
+                    if (preg_match('/AWG/', $name . $unit) === 1 && $number > $number2) {
+                        $tmp = $number;
+                        $number = $number2;
+                        $number2 = $tmp;
+                    }
+                    
                     return new self(name: $name, value_text: $value_text2, value_min: (float) $number,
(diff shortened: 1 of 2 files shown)


jbtronics:
Hmm i am not really a big fan of introducing a hardcoded special case for AWG. Something more universal would be good.

d-buchmann:
I see, and I'm all for a universal solution. But parameters themselves are very generic, so Part-DB basically knows nothing about them from the start.

I can think of two ways to approach this:
1) Exert more control
Put up a predefined list of parameters and attach special rules if needed (e.g. swapping min/max for reciprocal definitions like AWG or the "Conductance" of a resistor: $G = 1/R$). Then we can apply them if they match any parameter that is added.

2) Exert less control
Do not validate the numerical parameter values at all. This would make things much simpler, but then user typos or nonsensical responses from info providers could not be prevented.

Possible middle ground: Show a warning instead of an error on validation failure and allow saving anyway.
In my experience, this occurs rather seldom, so the user can fix it quickly or leave it as is, but doesn't have to sift though all parameters manually to check for plausibility.

What do you think would be the way to go?

PR #1413, opened 2026-06-15
Change part merge behaviour on info provider update

### LINKED ISSUE #1394: Editable info provider tag
It would be very useful if the Created by Information Provider tag were editable.
For example, when a passive component (like a resistor) becomes obsolete or cancelled, I often need to update the information provider and continue using the same part entry. Currently this is not possible, which complicates the workflow.

d-buchmann:
Fixes #1394
This changes the info provider "tag" if the user updates the part from a different info provider than it was created from, so favoring the more recently queried source by default.

### DIFF
--- a/src/Services/EntityMergers/Mergers/PartMerger.php
+++ b/src/Services/EntityMergers/Mergers/PartMerger.php
@@ -88,9 +88,9 @@ class PartMerger implements EntityMergerInterface
             return $t;
         }, $target, $other, 'manufacturing_status');
 
-        //Merge provider reference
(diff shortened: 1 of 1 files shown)


jbtronics:
not sure if this is still required after the info provider data of an part is editable.

d-buchmann:
Right, it's not required, but it may still be an enhancement.
The advantage of this implementation is that no additional user interaction is required.
The key question is if it is meaningful to keep the old info provider data.
If not, I think this should be merged additionally to making it editable.

PR #1426, opened 2026-06-24
Perform a full translation of Simplified Chinese.

0x915:
Based on the latest *.en.xlf file.

Translated automatically using DeepSinkV4 based on its strange interpretation of the source code,
then manually reviewed and corrected line by line.


jbtronics:
thanks.

You can check it out on the demo instance.

PR #1436, opened 2026-07-05
Import: document EDA/KiCad columns, add regression tests, fix eda_invisible inversion

Sebbeben:
## Summary

While importing parts with KiCad/EDA columns I hit some rough edges around the EDA import fields. This PR
does three small things: **documents** the EDA import columns (they were undocumented), **locks in** the two
supported column forms with tests, and **fixes** one alias that stored the wrong value.

## 1. Document the EDA / KiCad import columns

The `eda_*` import columns weren't documented anywhere, which made the flat names hard to discover. Added a
section to `docs/usage/import_export.md` listing them, and noting that the importer accepts **both**:

- the flat form (`eda_kicad_symbol`, `eda_value`, … and short aliases like `kicad_symbol`), and
- the nested `eda_info.*` form used by the CSV/JSON export and by the shipped `part_import_example.csv`
  (`eda_info.kicad_symbol`, …).

## 2. Regression tests for both column forms

The nested `eda_info.*` form already works today (the CSV encoder un-flattens the dotted headers and the
`eda_info` embeddable is in the `import` serialization group), but it wasn't covered by a test — so an export →
re-import round-trip, and the EDA columns in the example CSV, could silently regress. Added to
`EntityImporterTest`:

- `testImportAcceptsNestedEdaInfoColumns` — `eda_info.kicad_symbol` / `…footprint` / `…reference_prefix` /
  `…value` fill the part's `EdaInfo`.
- `testImportAcceptsFlatEdaColumns` — the flat form and its aliases still work.

## 3. Fix: `eda_invisible` stored the inverse value

`eda_invisible` was mapped to `eda_visibility` by a plain key-rename, so `eda_invisible=1` set
`visibility=true` — i.e. it made the part **visible**, the opposite of what the name says. It's now handled
explicitly and inverted, so `eda_invisible=1` correctly hides the part. Covered by
`testImportInvisibleAliasInvertsVisibility`.

## Notes

No schema/behaviour change beyond the `eda_invisible` fix; the nested-column support was already present and is
only being documented and tested here.

### DIFF
--- a/docs/usage/import_export.md
+++ b/docs/usage/import_export.md
@@ -109,6 +109,21 @@ supplier is not specified, the price and supplier product number fields will be 
 * **`supplier_product_number`** or **`supplier_part_number`** or * **`spn`**: The supplier product number of the part.
 * **`price`**: The price of the part in the base currency of the database (by default euro).
 
+The following fields set the EDA / KiCad metadata of the part (see [EDA / KiCad integration](eda_integration.md)).
+These can be given either as the flat column names below, or in the nested `eda_info.*` form used by the
+CSV/JSON export and the [example file](#example-data) (e.g. `eda_info.kicad_symbol`), so a file exported from
+Part-DB can be re-imported without renaming its headers:
+
+* **`eda_kicad_symbol`** or **`kicad_symbol`**: The KiCad symbol of the part, e.g. `Device:R`.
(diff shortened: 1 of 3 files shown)


jbtronics:
Looks good. Thank you.

PR #1438, opened 2026-07-07
Clarify library addition as HTTP in KiCad documentation

Apfelwurm:
One Word documentation addition to specify adding the Kicad Library file as a HTTP library.


jbtronics:
thanks

PR #1444, opened 2026-07-14
Add per-parameter "visible in symbol" flag for KiCad EDA export

Dant-hw:
## Summary

Adds a per-part-parameter flag that controls whether an exported parameter is
shown **on the schematic symbol** in KiCad, i.e. the KiCad field's `"visible"`
attribute in the EDA HTTP-library API response.

Previously Part-DB always emitted parameter fields with `"visible": "False"`,
so a parameter exported to KiCad was never displayed on the symbol without the
user manually toggling it in KiCad. This adds first-class control over that.

## Two independent "visibilities"

The existing `eda_visibility` and the new `eda_symbol_visibility` do different
things and are deliberately independent:

| Flag | Controls |
|------|----------|
| `eda_visibility` (existing) | Whether the parameter is **exported as a field** at all |
| `eda_symbol_visibility` (new) | Whether that field is **visible on the schematic symbol** (`"visible": "True"/"False"`) |

The new flag only has an effect once the parameter is actually exported.

## Behavior

- `eda_symbol_visibility` is a **nullable tri-state** on part parameters,
  mirroring the existing `eda_visibility`.
- When it is `null`, the new system default
  `KiCadEDASettings::defaultParameterSymbolVisibility` applies.
- Both default to `false`, so **existing installations keep the previous
  behavior** — exported parameter fields stay `"visible": "False"` until the
  flag (or the setting) is turned on. Fully backward compatible.

## Changes

- **Entity** – new nullable `eda_symbol_visibility` column on `AbstractParameter`
  with accessors and matching serialization groups; multi-platform migration
  (MySQL / SQLite / PostgreSQL).
- **Form** – a second `TriStateCheckboxType`, shown for part parameters only.
- **KiCadHelper** – resolves the flag (explicit value, else the system default)
  and passes it to `createField()`.
- **Settings** – `defaultParameterSymbolVisibility` (default `false`).
- **Templates** – new eye-icon column in the part edit *Specifications* table,
  next to the existing EDA-export column.
- **Translations** – English strings for the new tooltip and settings entries.
- **Tests** – assert `"visible": "True"` when the flag is set, and that the
  default stays `"False"`.

## Screenshot
partdb-EDA-visibility

## Testing

- `KiCadHelperTest` passes (24 tests, 45 assertions), including the two new cases.
- Migration applies cleanly on SQLite; the `eda_symbol_visibility` column is
  created as expected.
- Smoke-tested in the Docker image: container boots, migration runs on startup,
  the new column is present, and the edit form renders the new column.

## Notes

- Scope is **part parameters only** — orderdetails/supplier fields are not
  touched (could follow the same pattern later if wanted).
- Only `messages.en.xlf` is updated; other locales fall back to English until
  Crowdin syncs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/migrations/Version20260713120000.php
+++ b/migrations/Version20260713120000.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use App\Migration\AbstractMultiPlatformMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+final class Version20260713120000 extends AbstractMultiPlatformMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add eda_symbol_visibility nullable boolean column to parameters table (controls the "visible" flag of the exported KiCad field)';
+    }
+
+    public function mySQLUp(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE parameters ADD eda_symbol_visibility TINYINT(1) DEFAULT NULL');
+    }
+
+    public function mySQLDown(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE parameters DROP COLUMN eda_symbol_visibility');
+    }
+
+    public function sqLiteUp(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE parameters ADD COLUMN eda_symbol_visibility BOOLEAN DEFAULT NULL');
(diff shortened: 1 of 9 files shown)


jbtronics:
Looks good. Thank you.

PR #1447, opened 2026-07-18
TME info provider: migrate to APi verion 2

carmisergio:
### Problem
As per the [API documentation](https://api-doc.tme.eu), the v1 (Legacy) API will only be maintained until June 30, 2027. In addition, API tokens generated after May 14, 2026 can **only be used with API v2**, rendering it impossible for new users of Part-DB to access the TME info provider.

----

### Solution
Update the relevant `TMEProvider` implementation to make use of the new API. The structure of the new version is not very different from the legacy one, so no huge changes are necessary.
The only significant change is in how authentication is handled: the new API requires a specific call to the `/auth/token` endpoint, which returns a session token to be used for subsequent queries. The `info_provider.cache` is used to store the session token. The `TMEClient` transparently handles  requesting and renewing the session token.

#### Footprints
I have noticed that when both `Case - mm` and `Case -inch` (respectively IDs 2931 and 2932) are present in the parameter list, no `Case` (ID 35) is provided. In this situation, by default the provider uses the inch version to populate the footprint field, and I have added a new setting option to prefer the metric version instead.

---

### Testing
Added new tests for `TMEClient` to verify the new authentication logic.
Modified `TMEProvider` tests to reflect new API version.
All tests pass locally.

### DIFF
--- a/src/Services/InfoProviderSystem/Providers/TMEClient.php
+++ b/src/Services/InfoProviderSystem/Providers/TMEClient.php
@@ -39,7 +39,8 @@ class TMEClient
         private readonly TMESettings $settings,
         #[Autowire(service: 'info_provider.cache')]
         private readonly CacheInterface $cache
-    ) {
+    )
+    {
 
     }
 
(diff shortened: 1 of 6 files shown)


carmisergio:
Translation file was messed up, sorry about that. Should all be fixed now.

jbtronics:
Thanks for the PR. 
I had my own almost finished implementation of the TME v2 API. I will probably merge yours with my changes somehow.

But i will have to think about what is easier to use as an base, because both versions have some advantages over the other.

So could be that i will not merge this PR, but apply the changes in some other way.

jbtronics:
I merged my own implemenation.
Still  thank you very much for your PR, i took some inspirations from it.

carmisergio:
That's great. I made these changes just because I needed to import a  number of parts from TME and I figured this would take less time than doing it manually; then I realised it would make sense to contribute it back upstream.

Thank you for maintaining such a useful piece of software!

PR #1459, opened 2026-07-24
Consider the prices of built instances of subprojects

### LINKED ISSUE #1457: Associated builds part price
Hello!
It would be great if, when creating a part from a project (using Associated builds part), its price could be automatically taken from the project’s "Total build price" tag. This way, when creating nested projects, it would be possible to generate a multi‑level cost breakdown for the entire assembly.

d-buchmann:
This way the total price of a project sums up correctly when it contains subprojects with "built" parts (even if those parts have no explicit shopping information, but a BOM cost). Fixes #1457


jbtronics:
Looks good. Thanks.

PR #1465, opened 2026-07-28
Allow alternative names with commas

### LINKED ISSUE #1349: Comma in "Alternative Names" for categories prevents correct mapping
**Description:**
When using the "Alternative Names" field in category properties (Advanced tab), Part-DB uses a comma as a hardcoded separator to create a list of names.

**Problem:**
Many external information providers (e.g., DigiKey) use commas within their category names (e.g., "Trimmers, Variable Capacitors"). If this full string is entered as an alternative name, Part-DB splits it into two separate entries ("Capacitors" and "Ceramic"), which breaks the automatic category mapping during import.

**Expected Behavior:**
There should be a way to escape commas or use an alternative delimiter (like a semicolon) so that category names containing commas can be mapped correctly.

d-buchmann:
Simply replace the commas read from the info provider with wildcards, so they'll match any placeholder character the user chose in the alternative name definition.
Fixes #1349


jbtronics:
Its more a patch of an deeper issue, but as a workaround its okay I think.

I will think of a better solution for this for a deeper fix.

PR #1490, opened 2026-08-14
Add VS Code dev container support

alangibson:
Adds support for VS Code dev containers.

Dev container additionally exposes HTTPS port at 8443.


jbtronics:
thanks

PR #1495, opened 2026-08-19
Bump mcp/sdk to v0.7.1 (fixes CVE-2026-53965 SSE buffer DoS)

Sebbeben:
## Problem

The Symfony security checker (and `composer audit`) flag **`mcp/sdk` `v0.7.0`** as vulnerable:

- **CVE-2026-53965** / GHSA-7m52-jw36-44r3 — the client `HttpTransport` SSE buffer can grow unbounded when the server withholds the event delimiter, allowing a denial of service. Affected range: `>= 0.5.0, <= 0.7.0`.

This currently fails the `Static analysis` job's "Check dependencies for security" step on master (and consequently on every open PR).

## Fix

Bump the pinned `mcp/sdk` to **`v0.7.1`** (which bounds the SSE buffer), keeping the existing `as v0.6.0` alias. Only the `mcp/sdk` lock entry (and the content-hash) changes — no other dependencies are touched, and `composer update` reports *"No security vulnerability advisories found."*

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -74,7 +74,7 @@
         "league/oauth2-server-bundle": "^1.2.1",
         "liip/imagine-bundle": "^2.2",
         "maennchen/zipstream-php": "2.4.0.1",
-        "mcp/sdk": "v0.7.0 as v0.6.0",
+        "mcp/sdk": "v0.7.1 as v0.6.0",
         "nbgrp/onelogin-saml-bundle": "^v2.0.2",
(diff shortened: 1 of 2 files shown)


jbtronics:
Thanks.

There is however no security impact to Part-DB, as the vulnerability is used the client component of the mcp sdk, which is not used by Part-DB. Part-DB only uses the server version 

But it fixes the static analysis workflow.

PR #1496, opened 2026-08-19
Add Web NFC support for writing and scanning NFC tags

alangibson:
- Add NFC scanning to the existing barcode scanner page
- Add NFC tag writing to the part tools section
- Protect existing tag contents by default and require explicit overwrite confirmation
- Hide NFC controls on unsupported or insecure clients

Web NFC currently only works on Chrome on Android  when served over HTTPS.

### DIFF
--- a/assets/controllers/pages/barcode_scan_controller.js
+++ b/assets/controllers/pages/barcode_scan_controller.js
@@ -21,6 +21,7 @@ import {Controller} from "@hotwired/stimulus";
 //import * as ZXing from "@zxing/library";
 
 import {Html5QrcodeScanner, Html5Qrcode} from "@part-db/html5-qrcode";
+import { generateCsrfToken, generateCsrfHeaders } from "../csrf_protection_controller";
 import {decodeNdefMessage, isWebNfcAvailable, setScanInputAndSubmit} from "./nfc_helpers";
 
 /* stimulusFetch: 'lazy' */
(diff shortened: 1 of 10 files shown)


jbtronics:
Hey thanks for the PR.

What is the usecase for this?

Users can write the scan link of Part-DB onto an NFC tag, so they can be opened with a tap?
I guess there are no parts with manufacturer or distributor NFC tags.

alangibson:
Thanks for the great project. I'm dedicating a significant part of my life to organizing my disaster zone of an electronics lab and this is exactly what I need. 

My use case is affixing NFC tags to individual containers in assortment boxes. These boxes are already labeled with what's in them. See this: https://makerworld.com/en/models/769441-label-generator-gridfinity-bin?from=search#profileId-705079

It is possible to put a sticker with barcode on the side of the container. However, being 3d printed, the stickers don't stick to sides well. I'm pretty sure the will fail over time as individual boxes are moved around.

The bottoms are smooth and stickers stick well, but barcodes are inconvenient there. Hence I went with stick on NFC tags.

jbtronics:
Looks quite good in general:

A few things:
* For new controllers, translation should happen with the symfony ux frontend translator. So do not pass translation strings in templates, but move the translations to the frontend domain and use the trans() javascript function to load them directly in javascript. That makes the templates much less cluttered, and the controllers are more decoupled from the HTML.

* In most cases you do not want to create labels for a **part** (representing the part at locations), but instead linking to a specific **part lot**, which is a specific amount of things at a specific place. It would be good if you can write the lot specific urls to a NFC tag too

* The permission check in PartController for the `@labels.create_labels` is unnecessary. The scan URLs are not secret and you can easily construct them from the part ID.

PR #1498, opened 2026-08-20
Fix attachment merging

d-buchmann:
Allow auto-merging attachments with the same external path (if they have one), disregarding the internal path.

Until now, if there is a part with attachments from an info provider and you chose to download them, when you update the part from the info provider these attachmets get duplicated and you can't save unless you delete the duplicates by hand.


jbtronics:
Seems useful.
Thanks.

PR #1500, opened 2026-08-21
Move stock to new location

calebwest-SS:
This addresses Discussion #1247 to make stock management much simpler.
image
There is now an added radio button to "add stock" (using the same translation attribute as the "Add stock" button on the main page) that when selected allows the user to choose a location to move stock to, and creates a new stock lot when submitted.

The move lot button is now active on any stock lot that has non-zero quantity, as now there is functionality for parts that only have one stock lot.


jbtronics:
Looks great. Thanks

PR #1519, opened 2026-09-03
Add TrustedParts.com (ECIA) info provider

killecaptron:
TrustedParts.com is a part search engine operated by the Electronic Components Industry Association, which aggregates stock, prices and specifications of authorized distributors. Its Inventory API is free to use, but requires an approved account (company ID + API key).

The API has no part IDs, parts are always identified by their part number, so the provider IDs are built as "manufacturer|mpn". A single search already returns all offers and specifications, therefore the results are cached and reused for the detail view.

As the API can search up to 50 part numbers in a single request, the provider also implements BatchInfoProviderInterface for the bulk import.

Requests are sent with a user agent identifying Part-DB and the current user ID, as required by the API terms of use for non-website applications.

The datasheet links of this provider are redirect URLs without a file extension, which uncovered a separate problem in `Attachment::isPicture()` (broken preview images). That is fixed in #1520, independently of this PR.

### DIFF
--- a/docs/usage/information_provider_system.md
+++ b/docs/usage/information_provider_system.md
@@ -331,7 +331,43 @@ Once you have the API key, you can configure the Canopy provider in Part-DB usin
 
 * `PROVIDER_CANOPY_API_KEY`: The API key you got from Canopy (mandatory)
 
+### TrustedParts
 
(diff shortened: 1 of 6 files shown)


jbtronics:
Due to a lack of an API key i could not test it, but I assume you did.

And from the pure codebase it looks fine.

Thank you very much for the PR.

PR #1520, opened 2026-09-03
Fix broken preview images for extension-less external files

killecaptron:
If no file extension can be extracted from the URL of an external attachment, Part-DB assumed that it is a picture. That is a sensible guess for image URLs, but it also applies to URLs which redirect to the actual file, like the datasheet links of the TrustedParts provider (https://www.trustedparts.com/productredirect?id=...).

Such datasheets were rendered as an  in the attachment lists (showing a broken image) and the first one was automatically used as the preview picture of the part by the AttachmentSubmitHandler, so the part page showed a broken main image too.

The attachment type already knows which filetypes it may contain (the "Datasheet" type created by the info provider system is restricted to application/pdf), so use that information for the guess. Attachments whose URL has a known picture extension are unaffected, as are attachment types without a filetype filter.

Existing parts heal themselves: the AttachmentSubmitHandler already removes a master picture attachment which is not a picture anymore, as soon as the part is saved the next time.

### DIFF
--- a/src/Entity/Attachments/Attachment.php
+++ b/src/Entity/Attachments/Attachment.php
@@ -289,10 +289,43 @@ abstract class Attachment extends AbstractNamedDBElement
             //Check if we can extract a file extension from the URL
             $extension = pathinfo(parse_url($this->getExternalPath(), PHP_URL_PATH) ?? '', PATHINFO_EXTENSION);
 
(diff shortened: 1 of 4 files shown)


jbtronics:
Thanks for the PR. Seems like a useful change.

I moved the "allowsPicture" logic to the AttachmentType class, as its more a property of them and the logic might be reusable in other places.

PR #1530, opened 2026-09-09
Version the cache keys of the info provider DTOs

killecaptron:
The info provider system caches the DTOs it received from a provider as serialized objects, in a cache pool which is not part of the cache directory and therefore survives an update of Part-DB.

Whenever a DTO class gains a property, unserializing an object which was cached by an older version leaves that property uninitialized, so the first access to it fails with a typed property error until the cache happens to expire (up to four days later).

Add a version marker to every cache key of a DTO, which has to be increased whenever the structure of the DTOs changes: old entries are then simply never read again and expire on their own.


jbtronics:
Looks good. Thanks.

PR #1531, opened 2026-09-09
Fix duplicated attachments and unnecessary download attempts for tracker links

killecaptron:
Two issues which both surface when a part is updated from an info provider that serves its
files through tracking redirects, as TrustedParts does for every datasheet.

First, attachments were duplicated on every update. The merger only considered two attachments
identical if their external path matched as well, so a provider which issues a fresh signed or
tracking URL for the same file on every request produced a second attachment with the same name
and attachment type - which cannot be persisted at all, because PartAttachment enforces that
name and type are unique per part, so the update failed. Name and attachment type now decide,
which is exactly what that constraint already says, and a changed external URL refreshes the
existing attachment instead of adding a colliding one, so a stale or expired link does not
linger either.

Second, "download to server" was pre-selected for files which can never be downloaded. Those
tracking redirects only resolve for an actual browser and reject anything else (and the
TrustedParts terms of use prohibit downloading from their site anyway), so with
app.attachments.download_by_default enabled, every save ended in a download error. FileDTO
gained a downloadable flag which TrustedPartsProvider sets to false for its datasheet links,
and it is passed to AttachmentFormType as a form option, so all that changes is the default
state of a checkbox - nothing about it is persisted.

One thing worth pointing out for the review: the merge change applies to merging two parts as
well, not only to provider updates. Two attachments which share name and attachment type are
now merged into one, and the target's URL is refreshed from the other. Given the unique
constraint the previous behaviour could not have produced a valid result there either, but it
is a deliberate change rather than a side effect.

### DIFF
--- a/src/Form/AttachmentFormType.php
+++ b/src/Form/AttachmentFormType.php
@@ -174,7 +174,9 @@ class AttachmentFormType extends AbstractType
         //If the attachment should be downloaded by default (and is download allowed at all), register a listener,
         // which sets the downloadURL checkbox to true for new attachments
         if ($this->settings->downloadByDefault && $this->settings->allowDownloads) {
-            $builder->addEventListener(FormEvents::POST_SET_DATA, function (FormEvent $event): void {
+            $non_downloadable_urls = $options['non_downloadable_urls'];
+
+            $builder->addEventListener(FormEvents::POST_SET_DATA, function (FormEvent $event) use ($non_downloadable_urls): void {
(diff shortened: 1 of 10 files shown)


jbtronics:
Thanks for the PR.

I changed the merging behavior that it compares the canonical URLs without any query parameters or framgents, and use the other one if there is a difference.
That should keep the old behavior in most cases, while still fixing the issues with the changing query parameters of trustedparts

PR #1532, opened 2026-09-09
Allow refreshing parts from the info provider they were created with

killecaptron:
### Problem

Updating a part from its info provider means going through the provider search again, even
though the part already stores which provider and which provider-specific ID it came from. For
a part that is refreshed regularly this is three steps to reach a page the part could link to
directly — and there is no way at all to refresh several parts without repeating those steps
for each one.

### Change

Two steps, one per commit:

1. If a part was created by a provider which is still installed and active, the part page shows
   a button that goes straight to the existing update route with the stored provider key and ID,
   bypassing the search. The tooltip shows when the part was last updated from the provider.
2. The same for several parts at once, from the parts list. This reuses the existing bulk import
   infrastructure (`BulkInfoProviderImportJob`) rather than adding a second mechanism, so batch
   size, the part count limit and the timeout handling all apply unchanged.

The new `BulkRefreshResultsBuilder` sorts the selection out beforehand: parts which were not
created from a provider, and parts whose provider is no longer installed or is disabled, are
skipped and reported in the result instead of failing the whole job.

Both use `no_cache`, since the point of an explicit refresh is to get current data.

### Notes on rate limits

Every refreshed part costs one provider request, and providers enforce limits per account
(TrustedParts, for example, allows 50 requests per 10 seconds) and often bill per request. This
PR does not add any pacing of its own — it inherits the bulk import's batch size and part count
limit, which bound the work but not the request rate. I have a follow-up which paces Part-DB's
provider requests centrally; happy to submit that first if you would rather have the guard rail
in place before this.

### Permissions

No new permission: the button appears under the same conditions as the existing "update from
info provider" entry (`edit` on the part plus `@info_providers.create_parts`).

### DIFF
--- a/src/Controller/BulkInfoProviderImportController.php
+++ b/src/Controller/BulkInfoProviderImportController.php
@@ -31,6 +31,7 @@ use App\Entity\UserSystem\User;
 use App\Form\InfoProviderSystem\GlobalFieldMappingType;
 use App\Services\EntityMergers\Mergers\PartMerger;
 use App\Services\InfoProviderSystem\BulkInfoProviderService;
+use App\Services\InfoProviderSystem\BulkRefreshResultsBuilder;
 use App\Services\InfoProviderSystem\DTOs\BulkSearchFieldMappingDTO;
 use App\Services\InfoProviderSystem\DTOs\BulkSearchPartResultsDTO;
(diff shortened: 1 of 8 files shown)


jbtronics:
Sounds like a useful feature. 

Thanks.

Issue #1502, opened 2026-08-24
KiCad 10 ERC doesn't complain about local schematic field changes (request for approval)


mMuck:
**Describe the bug**
I'm a bit unsure with this and assuming this is a KiCad bug but I want to get this assumption approved from Part-DB users (because they have all requirements to approve this). A perhaps present limitation of HTTP libs could also be the source of this. Who knows(?). 🙂
 
I'm evaluating Part-DB as replacement for the [KiCad Database Library](https://docs.kicad.org/10.0/en/eeschema/eeschema.html#database-libraries) feature. As this is an atomic part use one important thing is that changes to local values/fields must be detected by the schematic ERC. This is not "firing" always. Investigating. Changes to (some?) fields in Part-DB will trigger the ERC. (Timestamp based or field by field compare?)
Updating symbols in KiCad loads the new values but bulk updating schematics is a bit risky - this should be detected, evaluated and done part by part. There is a diff dialog in KiCad.

With the Database Library this worked as intended (when I tested it 🤔). A local changed field (in the schematic) got detected. A meanwhile changed field in the DB got detected. This was live, no caching. Very secure not to export wrong/outdated data into a BOM.

**To Reproduce**
1. Add a part from a Part-DB lib into a schematic.
2. Prove that the ERC is clean.
3. Change the part value or a field like MPN (which should be updated always in the lib source => Part-DB)
4. Check the ERC again.
5. ERC isn't complaining.

**Expected behavior**
ERC should complain about changes in every field which is imported from Part-DB.

**Server Side**
- Part-DB Version: Part-DB 2.15.0 (docker compose installation)
- PHP Version: PHP 8.4.24
- Database Server: MySQL 8.0.46

**Desktop**
 - OS: Win11
 - KiCad 10.0.5 (latest stable)

PR #1533, opened 2026-09-09
Add settings to select which field groups are exported to KiCad

lukas-runge:
> [!NOTE]
> 🤖 Claude Fable 5.1 opening this PR on behalf of @lukas-runge

> [!IMPORTANT]
> Stacked on #1534 (cache fingerprint fix). Only the second commit belongs to this PR.

## Motivation

KiCad compares the fields of every placed symbol with the HTTP library and reports each difference as a "library symbol mismatch" in the ERC. Part-DB exports stock, storage locations, supplier part numbers, KiCost fields, part info and tags as symbol fields, so every stock booking, supplier edit or tag change in Part-DB produces ERC warnings in all schematics that use the part. In a workshop with regular stock movements this makes the ERC effectively unusable.

Related: #536 asked for a configurable field selection (in the other direction, more fields), #1502 discusses which field changes the ERC picks up.

## Changes

Five new settings in "KiCAD integration" (also via env, `EnvVarMode::OVERWRITE`):

| Setting | Fields |
|---|---|
| `EDA_KICAD_EXPORT_STOCK_FIELDS` | `Stock`, `Storage Location` |
| `EDA_KICAD_EXPORT_SUPPLIER_FIELDS` | `<Supplier> SPN` |
| `EDA_KICAD_EXPORT_KICOST_FIELDS` | `manf`, `manf#`, `<supplier>#` |
| `EDA_KICAD_EXPORT_PART_INFO_FIELDS` | `Category`, `Manufacturing Status`, `Mass`, `Part-DB IPN`, `Part-DB Footprint`, `Part-DB Unit`, `Part-DB Custom state` |
| `EDA_KICAD_EXPORT_TAGS_AS_KEYWORDS` | symbol keywords |

All default to `true`, so existing installations export exactly the same fields as before. Symbol, footprint, reference, value, datasheet, description, Manufacturer, MPN, Part-DB ID and Part-DB URL are always exported.

## Tests

- 7 new tests in `KiCadHelperTest` covering defaults, each switch, the independence of the supplier and KiCost switches, and that the new settings are part of the listing cache fingerprint from #1534.
- Docs: new section "Exported symbol fields" in `docs/usage/eda_integration.md`.

Running in production on a 2.17.0 instance since the 2.16 line.


jbtronics:
Looks good. Thanks.

PR #1534, opened 2026-09-09
Fix stale KiCad category listing after EDA settings change

lukas-runge:
> [!NOTE]
> 🤖 Claude Fable 5.1 opening this PR on behalf of @lukas-runge

## Problem

The KiCad category parts listing (`/kicad-api/v1/parts/category/{id}.json`) is cached in `KiCadHelper::getCategoryParts()`. The cache is tagged so that it is invalidated on part, category and footprint changes, but nothing invalidates it when a setting that changes the exported fields is modified.

To reproduce:

1. Let KiCad load a category, so the listing is cached.
2. In the server settings under "KiCAD integration" toggle "Datasheet field links to PDF" or one of the default visibility options for parameters/orderdetails.
3. Request the listing again. It still contains the old fields until some part in the database is edited.

There is no existing issue for this, I found it while working on #1533.

## Fix

Include a fingerprint of the settings that affect the serialized part (`datasheetAsPdf`, `defaultOrderdetailsVisibility`, `defaultParameterVisibility`, `defaultParameterSymbolVisibility`) in the cache key. A settings change then results in a new key and a fresh listing. Stale entries expire with the pool as before.

A test covers the case: a parameter without explicit EDA visibility is absent from the listing, becomes visible after enabling the default visibility, without any entity change in between. The test fails without the fix.

#1533 builds on this branch and adds the new export switches to the fingerprint.

### DIFF
--- a/src/Services/EDA/KiCadHelper.php
+++ b/src/Services/EDA/KiCadHelper.php
@@ -145,7 +145,10 @@ final readonly class KiCadHelper
      */
     public function getCategoryParts(?Category $category): array
     {
-        $cacheKey = 'kicad_category_parts_'.($category?->getID() ?? 0) . '_' . $this->category_depth;
+        // Settings fingerprint keeps cached responses from surviving settings changes
+        // (tag-based invalidation only fires on part/category/footprint edits)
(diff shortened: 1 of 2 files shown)


jbtronics:
thanks.

PR #1536, opened 2026-09-10
Pace the requests Part-DB sends to the info providers

killecaptron:
Info providers limit requests per account, usually in short windows - TrustedParts, for example,
allows 50 requests per 10 seconds and 150 per minute - and many of them bill per request.
Part-DB can easily produce bursts above that: a bulk import, or an update of many parts, walks
through them as fast as the network allows with nothing in between.

Worse, the limit belongs to the account, not to Part-DB. A script maintaining parts through the
API uses the same credentials, so its requests and Part-DB's add up while neither side can see
the other's count. Coordinating the two by hand is the only remedy today, and it fails the moment
somebody clicks "update from info provider" while a job is running.

Part-DB now paces itself. Two sliding windows per provider are enforced in PartInfoRetriever,
inside the cache callbacks - the single point every provider request passes, and the only place
where a cache hit can be told apart from a request that really goes out. The counters live in the
shared cache pool, so web requests, console commands and background jobs all count against the
same budget.

Requests are delayed, not rejected: a bulk operation simply takes longer, which is what it wants.
Only if a single request would wait longer than the configured maximum does it fail, with HTTP 429
and a Retry-After header naming the seconds until the next free slot, so an automated caller can
tell "too fast, come back shortly" apart from a real error.

The defaults (35 per 10 seconds, 120 per minute, 30 seconds maximum wait) stay deliberately below
what providers permit: the remaining headroom is what keeps Part-DB and an external script sharing
the account from pushing each other over the real limit. Each window can be switched off by
setting it to 0.

Two limitations worth naming. The counter updates are not synchronized - symfony/lock is not a
dependency - so parallel processes can slightly under-count, which is why the configured limit
should stay below the provider's own. And a provider with an internal cache of its own consumes a
slot even when it answers from that cache, which errs on the safe side.

This is the follow-up I offered in #1532.

### DIFF
--- a/docs/usage/information_provider_system.md
+++ b/docs/usage/information_provider_system.md
@@ -388,3 +388,31 @@ To reduce the number of API calls against the providers, the results are cached:
 If you need a fresh result, you can clear the cache by running `php .\bin\console cache:pool:clear info_provider.cache`
 on the command line.
 The default `php bin/console cache:clear` also clears the result cache, as it clears all caches.
+
+## Rate limiting
+
+Info providers enforce rate limits per account, usually in short windows (TrustedParts, for example, allows 50
+requests per 10 seconds and 150 per minute), and many of them bill per request. Part-DB can easily produce
+bursts above that: a bulk import or an update of many parts walks through them as fast as the network allows.
+
+Part-DB therefore paces its own requests to each provider. The limits are configured under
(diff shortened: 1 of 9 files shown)


jbtronics:
Thanks for the PR

PR #1537, opened 2026-09-10
Add an optional color to the custom part states

killecaptron:
A custom part state is shown as a badge on the part page and as plain text in the part tables, and
every state looks the same. States are typically used for things which differ in urgency though -
"needs review" and "do not use" should not look alike at a glance.

A state can now carry an optional semantic color, chosen from a fixed list of the Bootstrap
colors. The list is a closed whitelist: no free-form CSS or hex value ends up in the database, and
the mapping to a class happens in exactly one place. A state without a color keeps its previous
appearance, so nothing changes for existing installations until someone picks one.

The badge markup lives in PartDataTableHelper rather than being repeated in the part table, the
BOM table and the sidebar template, and the same helper renders the custom state column which the
project BOM table gained here (hidden by default).

The first commit is unrelated to the feature but needed by it: SkippableItemNormalizer treated any
string as an IRI, including the scalar value of a backed enum nested inside a resource, which
swallowed the value instead of letting the enum normalizer handle it. It now only does so for
types which really are API resources.

### DIFF
--- a/migrations/Version20260903120000.php
+++ b/migrations/Version20260903120000.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use App\Migration\AbstractMultiPlatformMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+final class Version20260903120000 extends AbstractMultiPlatformMigration
+{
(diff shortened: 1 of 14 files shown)


jbtronics:
Thank you.

I have improved a few things.

PR #1538, opened 2026-09-10
Store the stock a supplier has for a part, as reported by the info providers

killecaptron:
Note: this builds on #1530 and contains its commit, since a new DTO field is exactly the case that
fix is about. Once #1530 is merged this branch rebases onto it and the commit disappears.

Every info provider reports how much stock a distributor has, and Part-DB throws that away. It is
one of the first things you want to know when deciding where to order, and today it means opening
the distributor's page for each offer.

An orderdetail now stores the stock the provider reported, together with the time it was retrieved
at. Both are always written as a pair: a number without its age is misleading, because provider
stock goes stale within days, and there would be no way to tell a value from this morning from one
from three months ago. For the same reason the field is not part of the orderdetail form - it is
provider data, and setAvailableAmount() takes the value and its timestamp together.

Read from Digi-Key (the stock of the selected packaging, falling back to the product's), Mouser,
Octopart, Canopy and TrustedParts. In the merger the newer value wins, unlike the other fields:
stock is volatile, so a fresher reading replaces an older one, while null - meaning "unknown" -
never overwrites a known value.

Shown in the purchase information of a part with the time it was read at, plus a column in the part
table and one in the project BOM, both hidden by default: the BOM one answers "can I build this
project right now", which is otherwise a lot of clicking.

The migration is multi-platform (MySQL, SQLite, PostgreSQL); doctrine:schema:validate reports no
difference beyond what master already reports.

### DIFF
--- a/migrations/Version20260907090000.php
+++ b/migrations/Version20260907090000.php
@@ -1,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use App\Migration\AbstractMultiPlatformMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+final class Version20260907090000 extends AbstractMultiPlatformMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add available_amount and available_amount_updated_at columns to the orderdetails table, to store the stock a supplier had for a part and when that was retrieved';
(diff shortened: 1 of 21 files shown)


jbtronics:
Looks good.
Thanks.

PR #1540, opened 2026-09-10
Make the AI extraction schema valid for strict structured outputs

killecaptron:
The schema which the AI extractor sends along with its request is declared as a strict one (`'strict' => true`
in `DTOJsonSchemaConverter`), but it does not follow the two rules a schema has to follow to be accepted as
such: every object must forbid additional properties, and must list all of its properties as required. A
provider of the OpenAI family therefore refuses the request outright:

    Invalid schema for response_format 'part_detail': In context=('properties', 'parameters', 'items'),
    'additionalProperties' is required to be supplied and to be false.
    ("type":"invalid_request_error", "code":"invalid_json_schema", provider_name: Azure)

Which means the AI extractor cannot work with those models at all - "Create part from URL" fails every time.
Going through a gateway makes it harder to see: OpenRouter answers with its own "Provider returned error" and
keeps the reason in a field which is not part of the exception, so the failure looks like a misconfiguration on
the user's side. (I only got at the message above with #1539; the two are otherwise independent.)

Rather than sprinkling the two rules over a nested array literal, they are applied to the finished schema by
walking it. The schema stays as readable as it is now, and a property added later cannot forget them.

Requiring every property is only acceptable because the schema already expresses an optional value by allowing
null in its type, which the second test pins down alongside the rules themselves.

Verified on a real instance: with this change the extraction runs through against `openai/gpt-5.6-luna` via
OpenRouter, where it failed on every attempt before.

### DIFF
--- a/src/Services/InfoProviderSystem/DTOJsonSchemaConverter.php
+++ b/src/Services/InfoProviderSystem/DTOJsonSchemaConverter.php
@@ -41,7 +41,7 @@ final class DTOJsonSchemaConverter
      */
     public function getJSONSchema(): array
     {
-        return [
+        $schema = [
             'name' => 'part_detail',
             'strict' => true,
             'schema' => [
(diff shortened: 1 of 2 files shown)


jbtronics:
Thanks.

I never tested it with openai models, seems google's models are less strict about the schema

PR #1546, opened 2026-09-12
fix(mcp): gate the read-only get_part_details tool with read permission

### LINKED ISSUE #1545: MCP single-part get/update returns Access Denied while part search succeeds (v2.17.0)
Environment:
- Part-DB 2.17.0
- Docker deployment
- MCP_ENABLED=1
- MCP_EDITING_ENABLED=1
- Claude Code via Streamable HTTP MCP
- Edit-level personal API token (level 2)
- Dedicated non-admin user

Problem:

Collection-level Part MCP operations work, but single-Part object operations are denied even though effective Part permissions are Allow.

Observed behavior:

- search_parts / advanced_search_parts: works
- get_part_preview_image for Part ID 1: works
- get_part_details for Part IDs 1 and 2: Access Denied
- update_part for Part ID 1: Access Denied
- get_category_details: works
- list/get operations for categories, footprints, measurement units, manufacturers, suppliers, storage locations, projects: work

Part-DB CLI reports effective permissions:

Part / View   = Allow
Part / Edit   = Allow
Part / Create = Allow

Parts Stock / Withdraw = Allow
Parts Stock / Add      = Allow
Parts Stock / Move     = Allow
Parts Stock / Stocktake= Allow

API / Access API                    = Allow
API / Use MCP tools (for AI agents) = Allow

The Edit token is accepted normally:

GET /api/tokens/current -> 200 OK
token level = 2
user = partdb-agent

Claude Code reports the MCP endpoint connected.

Source observation:

In v2.17.0 Part.php, get_part_details appears to use:

security: 'is_granted("edit", object)'

even though it is a read-only MCP tool.

By comparison:

- search_parts uses @parts.read
- get_part_preview_image uses @parts.read
- get_category_details works normally

This looks like the MCP object-level security expression may be evaluated against the wrong subject before GetPartByIdProcessor resolves the actual Part, or there may be a related Part voter issue.

Additional possible bug:

While troubleshooting, the web permissions editor showed API permissions as enabled, but:

partdb:user:permissions partdb-agent

reported API Access / MCP as inherited and denied.

Setting those permissions via the CLI fixed API access immediately.

There may therefore also be a separate permission-editor persistence issue.

Happy to provide logs or additional tests.

wangzhengzhuo05:
## What

`get_part_details` is a read-only MCP tool, but its `McpTool` attribute was declared with the *mutating* permission expression `is_granted("edit", object)` — the same one used by `update_part`.

## Why it was broken

ApiPlatform/MCP enforces a tool's `security` expression on `tools/call` via `ExpressionAccessChecker` + `ResourceAccessChecker`. For an MCP tool there is no subject object bound, so `is_granted("edit", object)` can never resolve a `Part` and always denies with "Access Denied", while the sibling read-only tools (`search_parts`, `advanced_search_parts`, `get_part_preview_image`) that use `is_granted("@parts.read")` work fine.

That matches the report in #1545 exactly: search and the preview image work, `get_part_details` is denied even for a user whose effective Part/View and Part/Edit permissions are both Allow.

The per-object check is not lost by this change: `GetPartByIdProcessor` already performs

```php
if (!$this->authorizationChecker->isGranted('read', $part)) {
    throw new AccessDeniedException(sprintf('Access denied to part with id %d', $data->id));
}
```

before returning the part — the identical check its working sibling `GetPartPreviewImageProcessor` uses. The declared tool security only needs to express the collection-level gate, which is why every other read-only tool on this resource uses `@parts.read`.

## How

One line in `src/Entity/Parts/Part.php`: `get_part_details`'s `security` is aligned with the other read-only tools → `is_granted("@parts.read")`. No other tool is touched — `create_part`, `update_part` and `delete_part` keep their mutating expressions.

## Tests

New `tests/State/Mcp/ReadOnlyMcpToolsUseReadSecurityTest.php` (KernelTestCase, same approach as the existing `CallToolResultToolsDeclareNoOutputSchemaTest`). It inspects the registered ApiPlatform metadata of the Part resource — the same metadata the MCP server enforces — and

* asserts the rule that every tool annotated `readOnlyHint: true` must not be gated by an `edit`/`create`/`delete` expression (this catches the whole class of bug, not just this instance);
* pins `get_part_details` to `is_granted("@parts.read")`.

Results (PHP 8.3.6, PHPUnit 11.5):

* `./bin/phpunit tests/State/Mcp/ReadOnlyMcpToolsUseReadSecurityTest.php` → OK (2 tests, 17 assertions)
* `./bin/phpunit tests/State/Mcp/` → OK (140 tests, 673 assertions)
* Mutation check: reverting only the one-line production fix makes both new tests fail with `-'is_granted("@parts.read")' / +'is_granted("edit", object)'`; restoring it turns them green again — so the test really guards this fix.

Fixes #1545

*(This change was implemented with AI assistance (OpenCode + muse-spark-1.3) and then reviewed, tested and mutation-checked by hand.)*

### DIFF
--- a/src/Entity/Parts/Part.php
+++ b/src/Entity/Parts/Part.php
@@ -154,7 +154,7 @@ use Symfony\Component\Validator\Context\ExecutionContextInterface;
                 'groups' => ['part:read', 'provider_reference:read', 'api:basic:read', 'part_lot:read', 'orderdetail:read', 'pricedetail:read', 'parameter:read', 'attachment:read', 'eda_info:read'],
                 'item_uri_template' => '/api/parts/{id}',
             ],
-            security: 'is_granted("edit", object)',
+            security: 'is_granted("@parts.read")',
             input: ElementByIdInput::class,
             validate: true,
             processor: GetPartByIdProcessor::class
--- a/tests/State/Mcp/ReadOnlyMcpToolsUseReadSecurityTest.php
+++ b/tests/State/Mcp/ReadOnlyMcpToolsUseReadSecurityTest.php
@@ -1,0 +1,119 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Part-DB (https://github.com/Part-DB/Part-DB-symfony).
+ *
+ *  Copyright (C) 2019 - 2026 Jan Böhmer (https://github.com/jbtronics)
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU Affero General Public License as published
+ *  by the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+namespace App\Tests\State\Mcp;
+
+use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
+use App\Entity\Parts\Part;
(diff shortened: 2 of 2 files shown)


jbtronics:
Thanks. 

In the earlier versions of api core the security wasnt actually enforced, so i guess i did not notice that mistake earlier.

PR #1552, opened 2026-09-19
Fix dynamic form collection reset and parameter uniqueness validation

LMatt08:
## Summary

This PR fixes two issues affecting dynamic form collections:

* Restore the initial collection structure when a form is reset, including rows removed or added after the initial page load.
* Properly unregister the TomSelect form-reset listener when an instance is destroyed.
* Ignore parameters no longer present in their owning Part collection during uniqueness validation. This allows removing a parameter and recreating it with the same name in a single form submission, while still rejecting two simultaneously active parameters with the same name.

These generic changes were extracted from #1492 following review feedback. This PR contains no parameter-definition, choice-list, filtering, migration, or permission changes.

## Implementation

The changes are split into two commits:

1. `Fix resetting dynamic form collections`

   * Restores the initial collection structure on form reset.
   * Cleans up the registered reset listener when TomSelect is destroyed.

2. `Ignore removed parameters during uniqueness validation`

   * Adds a dedicated repository method used by the `UniqueEntity` constraint.
   * Considers only parameters still present in the owning Part collection.
   * Adds focused regression tests independent of the parameter-definition feature.

## Verification

A negative-control run against the unmodified `upstream/master` reproduced the issue:

```text
Tests: 2, Assertions: 3, Failures: 1
testRemovedParameterCanBeRecreatedWithTheSameNameBeforeFlush
ERROR: This value is already used.
```

With the changes applied:

```text
Focused regression tests: 2 tests, 9 assertions — OK
PartParameterTest + PartControllerTest: 46 tests, 72 assertions — OK
PHPStan targeted analysis: OK
ESLint targeted analysis: OK
Yarn production build: OK
git diff --check: OK
```

The repository currently has no DOM/browser test infrastructure for the JavaScript reset behavior, so the modified JavaScript was verified through ESLint and the production asset build.


jbtronics:
Thanks for the PR.

I cherry picked the form reset commit. 
For the parameter uniquness validation, i implemented a far more generic approach, which implements a custom `#[UniqueEntityIgnoringOrphans(fields: ['name', 'group', 'element'], ownerField: 'element')]` validator, which just uses doctrine metadata to check that automatically, without the need of a custom repo function.
This can be easily applied to both parameters and attachments for all element types.

